PR #24, opened 2015-11-28
fix parsing of comments

Gibheer:
It looks like there is an error in the configparser to parse comments.
This results in item.name in the whisper_schema parser to end up with
the whole comment.

I will also try to fix the configparser, so that this isn't needed in the future, together with the other fix.


Gibheer:
Thank you. And another thank you for the good work!

PR #27, opened 2015-12-02
Check for NaN values when parsing line protocol

### LINKED ISSUE #17: NaN values being saved in Whisper datafiles
hi @lomik , With reference to https://bugs.launchpad.net/graphite/+bug/867882

Somehow I'm finding my whisper files to contain `nan`, and while this is valid from a python JSON spec, it's breaking the Grafana javascript code which parses the JSON from graphite. And causes an error and does not plot graphs. 

Apparently carbon-cache was patched to eliminate such cases of writing nan to the whisper files. Can you look into this?

ahayworth:
Go considers a 'NaN' string to be a valid Float ... but unfortunately
that's not actually something graphite-web or grafana knows how
to handle.

Much like the original carbon-cache, do not persist such values and
reject them.
https://github.com/graphite-project/carbon/blob/83b87729a98a7b8efb3f0119280da3f812614ea2/lib/carbon/protocols.py#L65

Fixes #17.

### DIFF
--- a/points/points.go
+++ b/points/points.go
@@ -71,13 +71,13 @@ func ParseText(line string) (*Points, error) {
 
 	value, err := strconv.ParseFloat(row[1], 64)
 
-	if err != nil {
+	if err != nil || math.IsNaN(value) {
 		return nil, fmt.Errorf("bad message: %#v", line)
 	}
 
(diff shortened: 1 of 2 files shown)


lomik:
Thanks!

PR #28, opened 2015-12-02
Explicitly test against go 1.5.1

lomik:


ahayworth:
Whoops - Sorry about that. I rebased that into a branch after I realized i'd pushed it onto the last pull.

In any case, I'm seeing intermittent failures against go1.5.1, so this should help catch them. Thanks! :)

PR #43, opened 2016-05-04
clean up schemas code a bit and export some types for other packages to reuse

Dieterbe:
- simplify code & remove some unneeded pointer referencing
- make the schema fields public, so that other projects can use them. I have a use case for this.
- remove logging statement from library code. logging is a higher level concern
- fix comments for exported types
- clear up some code flow issues
- fix a bad printf format


Dieterbe:
note: this requires https://github.com/lomik/go-whisper/pull/3

PR #72, opened 2016-08-01
Convert project to govendor

### LINKED ISSUE #71: Convert to use govendor for vendored libraries
This project predates govendor as a supported first party solution for vendoring in 3rd party libs and uses submodules instead. Now that golang has native vendor tooling/support it seems anachronistic to stay with the non-standard submodule setup.

drawks:
- This resolves #71 by relocating "_vendor/src" to "vendor" and adds a
  vendor.json file to peg initial values
- Ignore dirty submodules to prevent noisy git status


drawks:
Looks like this needs a bit more work to maintain backwards compat with go versions prior to 1.5

drawks:
Closing this PR for now while I sort out travis

PR #79, opened 2016-08-09
Optimize writeout queue in cache

redbaron:


drawks:
These look pretty rad. I love the "noop" write strategy, it was a huge perf gain not waiting for sorts in the original python carbon-cache.

I'm a little confused why these are all in one big PR though. 

42dc917 and 2aa0967 Seems like they should be a single PR and should have a benchmark to go along with it
126f00b Can stand on it's own and doesn't come with the inherent risk with changing the underlying data model/handling
 121cca0 Similarly low risk and can stand on it's own

At any rate, it isn't my project to manage. But these look really cool.

redbaron:
Agree, I'll split into  multiple PRs.

"noop" is the only way to go, all other "strategies" are redherring and do more harm than good

Any properly configured carbon server will be tuned so that go-carbon writes to pagecache only and would almost never be blocked on IO  (it can happen only  to read headers which were evicted from pagecache since last write, but that is not affected by order anyway), so there is no difference in which order to writeout datapoints.

deniszh:
:+1: for implementing 'noop' strategy!
But agreed with @drawks with splitting this.

redbaron:
I am preparing commit with a benchmark, expect updates soon

redbaron:
OK here are test results:

```
# checkout first commit in PR with benchmark added only
$ git checkout 9eb8ffd37c084106a83b5cc870e220ff4ae86b66
$ go test -v -run zzzzz -bench .  -benchtime=20s -benchmem
PASS
BenchmarkUpdateQueueMax-4         20    1508619480 ns/op    77187959 B/op    1000052 allocs/op
BenchmarkUpdateQueueSort-4        20    1723601365 ns/op    77187940 B/op    1000050 allocs/op
BenchmarkUpdateQueueNoop-4        30    1117548096 ns/op    77187950 B/op    1000054 allocs/op
ok      _/home/rossmohax/repos/external/go-carbon/cache 115.608s
```

So baseline is **1508 ms**, **1724 ms** and **1118 ms** for  all three strategies
With queue optimizations:

```
$ git checkout reuse-cache-queue 
$ go test -v -run zzzzz -bench .  -benchtime=20s -benchmem
PASS
BenchmarkUpdateQueueMax-4         50     636295082 ns/op          32 B/op          1 allocs/op
BenchmarkUpdateQueueSort-4        10    2793152131 ns/op          32 B/op          1 allocs/op
BenchmarkUpdateQueueNoop-4       500      49195900 ns/op           0 B/op          0 allocs/op
```

That is **636ms (2.33x speedup)**,  **2793ms (1.6x slowdown)**  and **49 ms (22x !! speedup)**

Slowdown for timestamp sort is due to multiple dereferencing to reach timestamp.

I have an attempt in my repo (https://github.com/redbaron/go-carbon/tree/queue-optimize-bad-sort) where I try to improve sorted strategy case and although it improves numbers comparing to `master`, it comes with a 10x  penalty comparing to fastest case presented here.

redbaron:
Another note: before PR benchmark memory floats ~1.1 GB on Go 1.6.3 , with changes in this PR it comes down to ~650MB

PR #81, opened 2016-08-10
Noop write strategy

redbaron:


redbaron:
Failing test is flaky

PR #84, opened 2016-08-11
Fair queue

redbaron:
If we reorder writeout queue before it is exhausted
and time to process queue is less than checkpoint interval
then it can cause less active metrics to never be written
to disk, always residing in a cache, because just before
they are about to be written new queue is built and processed
from the top again.

This is more visible if "max" write-strategy is used
and there are a lot of active metrics.

By removing `updateQueue()` call from checkpoint we make it
fair - queue once build is always processed fully before
it is built again. It guarantee that every single
metric will be written to disk.

Without `updateQueue` checkpoint is doing nothing more
than submitting stats, so some stats like 'checkpointTime' are
not that interesting anymore.

Metrics inputLenBeforeCheckpoint and inputLenAfterCheckpoint
were renamed to maxInputLenBeforeQueueRebuild and
maxInputLenBeforeQueueRebuild and contain worst values from
all `updateQueue` runs happen between checkpoints

### DIFF
--- a/cache/cache.go
+++ b/cache/cache.go
@@ -51,6 +51,8 @@ type Cache struct {
 	queue          queue
 	queueBuildCnt  int           // number of times writeout queue was built
 	queueBuildTime time.Duration // time spent building writeout queue
+	maxInputLenBeforeQueueRebuild int
+	maxInputLenAfterQueueRebuild  int
 }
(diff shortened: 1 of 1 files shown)


redbaron:
rebased

PR #91, opened 2016-08-14
Stat atomics helper and usage example

redbaron:


lomik:
helper.SendAndZeroUint32 can't be used for "maxInputLenBeforeQueueRebuild" and "maxInputLenAfterQueueRebuild"

redbaron:
True! updated PR

redbaron:
BTW, I am not sure that using atomics comes with neglible performance penalty, especially those which are updated per metrics. 

Since only 1 thread updates them, maybe there is away to avoid use of most of them, but for that I'd need to have better understanding of go memory model in regards to atomics,it doesnt seem to be documented at all, although there is a useful discussion here https://github.com/golang/go/issues/5045

lomik:
I tried different options here https://github.com/lomik/go-carbon/blob/master/cache/stat_test.go

If there is a noticeable slow down, you can use a solution of the test BenchmarkCounterAtomicAddWithDelta

PR #92, opened 2016-08-15
Fix queueWriteoutTime metric skew towards metricInterval value

redbaron:
Due to `.size == 0` check in .Pop() .Get() method wasn't called the
moment queue was exhausted, which delayed metric update up to
a metricInterval.

Moved `.size` check to updateQueue, which allows metric to be
updated precisely


redbaron:
also removed duplicate call in stat

PR #93, opened 2016-08-16
Use whisper.TimeSeriesPoint straight from the start

redbaron:
It avoids moving to heap every datapoint in persisters
and potentially allows to remove slice rebuild if go-whisper
is ever updated to receive slices of values, not pointers

### DIFF
--- a/cache/cache.go
+++ b/cache/cache.go
@@ -22,7 +22,7 @@ func (v byLength) Less(i, j int) bool { return len(v[i].Data) < len(v[j].Data) }
 
 func (v byTimestamp) Len() int           { return len(v) }
 func (v byTimestamp) Swap(i, j int)      { v[i], v[j] = v[j], v[i] }
-func (v byTimestamp) Less(i, j int) bool { return v[i].Data[0].Timestamp > v[j].Data[0].Timestamp }
+func (v byTimestamp) Less(i, j int) bool { return v[i].Data[0].Time > v[j].Data[0].Time }
 
 type WriteStrategy int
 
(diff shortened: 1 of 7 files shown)


lomik:
I do not want to go-carbon so strongly dependent on the data storage library. Perhaps an alternative storage backends will be in the future (eg https://github.com/graphite-project/ceres)

How noticeable improvement in performance after this change?

redbaron:
We are talking about `committedPoints` allocations per minute, which is not neglible even with constantly improving GC in Go runtime.

Similar change when`points.Points.Data` was changed from `[]*Points` to `[]Points`  knocked down average CPU time from 122% to 103% on our servers committing 3.5M datapoints/minute. 

I'll measure effect of this patch separately and come back with some numbers

redbaron:
In the meantime updated patch to use type alias, so that go-whisper is imported only once and rest of go-carbon operates on `points.Point` which just happen to match layout of `whisper.TimeSeriesPoint` ;)

Compatibility of structs is checked by compiler at a pointer casting time in `persister/whisper.go` so if you ever evolve `point.Point` into something else, compiler will detect it and fail compilation, at which point you might revert this optimisation

redbaron:
`pidstat -p <pid> 1 60`, run 3 times

before: shows 104-108 CPU usage  as 60 seconds average
after: shows 96-102 CPU usage  as 60 seconds average

lomik:
Is there a measurable increase in work speed (save to disk, carbonlink answers)? Or simply reduced CPU usage?

I do not think that a significant acceleration. Independence of the application modules from each other and from whisper to me is more important. I re-use some of the carbon modules (receivers) in other private projects that are independent of whisper.

PR #94, opened 2016-08-16
Fix extra dot after default graph prefix value

redbaron:
Proper way would be to change default and examples, but
that would require config update for all the users


lomik:
Let us go the right way. Users of the old config will not break anything. Metric "prefix..value" stored to "prefix//value.wsp" (same as prefix/value.wsp)

lomik:
Default changed here https://github.com/lomik/go-carbon/commit/ad8d13c967039b4208dad2c0e2860881f4268d8e . Readme after release

Issue #87, opened 2016-08-11
Spurious test failure seen in Travis


drawks:
```
?       _/home/travis/gopath/src/github.com/drawks/go-carbon    [no test files]
--- FAIL: TestCarbonlink-2 (0.05 seconds)
    Location:   carbonlink_test.go:115
    Error:      Not equal: "\x80\x02}(X\n\x00\x00\x00datapoints](J\xd3)\xceTG\xc0E\x11\xeb\x85\x1e\xb8R\x86J\xbe$\xceTG@.\x00\x00\x00\x00\x00\x00\x86eu." (expected)
                    != "\x80\x02}(X\n\x00\x00\x00datapoints](J\xd3)\xceTG\xc0E\x11\xeb\x85\x1e\xb8R\x86eu." (actual)

    Location:   carbonlink_test.go:139
    Error:      Not equal: "\x80\x02}(X\n\x00\x00\x00datapoints](eu." (expected)
                    != "\x80\x02}(X\n\x00\x00\x00datapoints](J\xbe$\xceTG@.\x00\x00\x00\x00\x00\x00\x86eu." (actual)

FAIL
FAIL    _/home/travis/gopath/src/github.com/drawks/go-carbon/cache  0.562s
```

While checking in something completely unrelated, I noticed this test error pop on travis for a go1.3 based build. Not sure if it may be related to #83 or not

lomik:
>  Not sure if it may be related to #83 or not

Most likely you're right. Perhaps this race was to blame https://github.com/lomik/go-carbon/commit/87068714a87493500227cb50679186ca17942046#diff-e648ab02e8f8bc645ee3db96ece8ca4bR39

But perhaps the reason is something else. Hard to say for sure, because I can't reproduce it

drawks:
I can reliably reproduce this error by running the cache tests in a loop:

`err=0; while [ ${err} = 0 ]; do go test -v -race ./cache ; err=$?; done`

`--- FAIL: TestCarbonlink (0.00s)
        Error Trace:    carbonlink_test.go:102
    Error:      Not equal: "\x80\x02}(X\n\x00\x00\x00datapoints](J\xd3)\xceTG\xc0E\x11\xeb\x85\x1e\xb8R\x86J\xbe$\xceTG@.\x00\x00\x00\x00\x00\x00\x86eu." (expected)
[...]

Issue #83, opened 2016-08-10
Tests indicate races in cache and receiver


drawks:
I'm not super familiar with the race detector in go, so I'm uncertain if this is indicative of a race in the test code vs a race in the library code being tested. Either way, I suspect it should be cleaned up and the race detector enabled in the "test" make target so it gets better coverage in travis.

```
$ go test -race ./cache
....

==================
PASS
Found 9 data race(s)
FAIL    github.com/lomik/go-carbon/cache    1.570s
```

```
$ go test -race ./receiver

....

==================
PASS
Found 19 data race(s)
FAIL    github.com/lomik/go-carbon/receiver 1.170s
```

drawks:
After poking at this for a little bit I think the race is actually in `logging.Test` which allows an unsynchronized bytes.Buffer to be written/read in different goroutines. I'm actually a little puzzled by how the logging package is envisioned to be used since I see the `logrus` package directly imported and used all over the place and even a smattering of the standard library `log` gets used during the initialization of carbon-agent.

FWIW I /think/ we should be initializing a logrus logger instance in the `lomik/go-carbon/logging` package and then providing that logger to all the various packages that want to log things.

It looks like `logrus` also provides some decorated logger instances designed as testing shims. I'm going to take a quick crack at converting the racy tests to use these facilities directly for now and perhaps try and do a logging overhaul in another issue/pr.

lomik:
Some races fixed here https://github.com/lomik/go-carbon/commit/87068714a87493500227cb50679186ca17942046

Nothing interesting, only tests code affected

```
$ go test -race ./cache
...
==================
PASS
Found 1 data race(s)
FAIL    go-carbon/cache 1.608s


$ go test -race ./receiver
ok    go-carbon/receiver 1.242s
```

There was only one problem

```
go test -race ./cache
==================
WARNING: DATA RACE
Read by goroutine 44:
  github.com/Sirupsen/logrus.(*Entry).Debugf()
...

Previous write by goroutine 39:
  github.com/Sirupsen/logrus.SetLevel()
...

==================
PASS
Found 1 data race(s)
FAIL    go-carbon/cache 1.602s
```

Race condition in logrus:  https://github.com/Sirupsen/logrus/issues/295
Also only in tests (really not affected, but auto detected). In normal operation, the logging level is made once at the start (before first log attempt from concurrent thread) and never changing.
There is a chance that the fix in the near future https://github.com/Sirupsen/logrus/pull/374

[...]

PR #98, opened 2016-08-24
More tests and conn handling improvements

drawks:
- Added a bunch more tests to carbonlink
- Removed tests for logging side effects in favor of direct isolated unit
  tests #83 #87 
- modified carbonlink to immediately drop connection on malformed queries
- Enabled race detector in Makefile's test target


drawks:
This PR should improve test coverage and reduce dependence on testing for exact language in errors and/or log messages.
The only behavioral change is that connections are dropped immediately on carbonlink queries that timeout while reading from the conn or fail on parsing the query packet. Both cases seemed like a sane behavior as there is no way to correct for receiving phase after an incomplete read.

lomik:
Please, don't enable "-race" by default before https://github.com/Sirupsen/logrus/issues/295 fixed.

drawks:
Tests should pass with race detection enabled, this PR handles the bug in logrus by removing calls to change the state of the logger so that bug doesn't trigger race detection.

cache/carbonlink.go
```diff
@@ -133,13 +133,15 @@ func (listener *CarbonlinkListener) handleConnection(conn net.Conn) {
 
 		reqData, err := ReadCarbonlinkRequest(reader)
 		if err != nil {
+			conn.(*net.TCPConn).SetLinger(0)
```
redbaron, line 136: shouldn't it call `conn.Close()` ? if message is malformed there are no sync markers in protocol to find beginning of new message, it all going to be garbage

lomik:
> Tests should pass with race detection enabled, this PR handles the bug in logrus by removing calls to change the state of the logger so that bug doesn't trigger race detection.

Ok

There is another problem. Race detector greatly slows down tests. On my MacBook started to fall another test

```
--- FAIL: TestThrottleChan (8.01s)
        Location:       throttle_test.go:38
    Error:      Should be true
    Messages:   perSecond: 531234, bw: 347932

FAIL
```

lomik:
In travis same problem

```
--- FAIL: TestThrottleChan (7.00s)
    Location:   throttle_test.go:38
    Error:      Should be true
    Messages:   perSecond: 213000, bw: 200221

FAIL
```

cache/carbonlink.go
drawks: The break after the message jumps out of the for loop which causes the rest of the function to run and then return at which point the deferred Close runs. If I've got the flow wrong here please advise.

drawks:
The throttle chan test should really run as two steps, one where it runs unthrottled to determine the throughput in the given test env followed by a throttled run which sets the throttle to some significantly reduced rate(20% of the max or similar).

cache/carbonlink.go
redbaron: true, missed it

PR #99, opened 2016-08-24
Added end-to-end benchmark test

redbaron:
To be used to measure future perf improvementss


lomik:
Tests is broken :(

redbaron:
After dozens of rebases forgot to chck state of other tests,I'll create a separate PR

PR #102, opened 2016-08-30
Reduce GC pressure in benchmark to reduce test-to-test variance

redbaron:


redbaron:
Failed test has nothing to do with this change

PR #103, opened 2016-08-30
Handwritten pickle-unpickle operations

redbaron:
Synthetic benchmark doesn't do it justice,  but tons of GC generated by `reflect` are limiting number CL requests in end-2-end benchmark by factor of two.

Synthetic differences you can see below

Before:

```
go test -bench "BenchmarkCarbonLink"  -run none .  -benchtime 10s -benchmem
PASS
BenchmarkCarbonLinkPickleParse-4    20000000           967 ns/op        1144 B/op          8 allocs/op
BenchmarkCarbonLinkPackReply-4       2000000          8957 ns/op        2688 B/op        120 allocs/op
```

After:

```
go test -bench "BenchmarkCarbonLink"  -run none .  -benchtime 10s -benchmem
PASS
BenchmarkCarbonLinkPickleParse-4    100000000          105 ns/op          96 B/op          1 allocs/op
BenchmarkCarbonLinkPackReply-4      30000000           458 ns/op         224 B/op          3 a
```

### DIFF
--- a/cache/carbonlink.go
+++ b/cache/carbonlink.go
@@ -90,7 +90,7 @@ func (listener *CarbonlinkListener) SetQueryTimeout(timeout time.Duration) {
 	listener.queryTimeout = timeout
 }
 
-func (listener *CarbonlinkListener) packReply(query *Query) []byte {
+func packReply(query *Query) []byte {
 	buf := new(bytes.Buffer)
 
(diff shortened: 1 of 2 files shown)


lomik:
I have long wanted to do this myself. Thanks!

drawks:
Nice change!

PR #107, opened 2016-09-02
Support unicode strings when unpickling

redbaron:


drawks:
@redbaron  I'm trying to resolve a few issues that are open which need pickled responses and with the new hand written pickle code it is a bit more daunting. Can you point me to whatever reference you are using for the pickle protocol?

redbaron:
I used /usr/lib64/python2.7/pickle.py

drawks:
Ah, I actually found python's pickletool library in the stdlibrary does a decent job of showing pickle disassembly.

PR #109, opened 2016-09-05
Parallel cache 

redbaron:
This series of patches refactors inner logic to remove bottleneck on  a single Cache goroutine. 

At it's core there is a ConcurrentMap (CMap)  which is a just an array of maps and corresponding  RWLocks.

All appropriate locking is localized in CMap itself (shard-level locking) and new `writeoutQueue` (metric level locking), rest of the code just uses API.

Receivers now add points straight into CMap. There are 128 shards, allowing up to 128 receivers to add points in parallel. Real level of parallelism achieved in end-to-end benchmark is ~12-15.

Persisters now receive batches to write in a form of  `[]*point.Point`.  They writeout however many points they see and then cut written points shifting remaining to beginning of `point.Point.Data` slice (see `writeoutQueue.Process` for details). There is no  need for confirm tracker anymore - `writeoutQueue.Process`  takes care of updating CMap.  There is a rudimentary strategy to keep entry around for 1 more writeout cycle, so that frequently updated metrics will not  cause rapid allocaion/deallocation cycle. Better strategies might be developed in the future. Without `inputChan` to overflow relevant tunables and stats were removed as obsolete.

CarbonLink handlers now query CMap directly. It slashes response times and make them more stable, no more "CarbonLink took more than 100ms" errors, querylink timeout tunable was removed.


redbaron:
I can't reproduce 1.6 and 1.7 travis-ci failures :(

redbaron:
`go vet` result for Go <1.7 is incorrect, 1.7 release notes reflect it too

lomik:
I expected that the cache may become a bottleneck, but did not believe that will happen in practice.
At what load it happened (pps, uniq metriсs count, whisper scheme, hardware)? 

I need a lot of time for explore your changes. Now I have no time, so I can't promise a quick decision.

At first glance, I do not like what is lost modules isolation. This can be avoided, if pass callback function to receivers (not cache instance).

redbaron:
With 3M points/minute and 7k  carbonlink request/minute, requests are often timed out even with timeouts as high as 1 second given that we alert on graphite in real time empty responses from timeout carbonlink requests are critical for us.

Also weekly we had cases when go-carbon rapidly reaches limit of filedescriptors and then spams gigabytes of logs – this behaviour might be due to bug and not performance problems, but it's gone with one of the previous versions of this patchset. I had no chance to investigate previous misbehaviour, so can't provide even lsof for details ☹

When IO is a bottleneck (for instance when starting with a cold pagecache), it can be seen how this version can drive 64 writers easily (60-62 locked in D state that's what I’d expect), previous current master can do only 13-15 persisters at a time, rest are waiting for cache goroutine. SSDs throughput grows somewhat close to linear with queue depth up until ~256 QD, so being able to issue enough IO in parallel is important.

End to end benchmark shows results twice as high, I’ll retest and post.

performance was a driving factor for a patchset, but as a result code actually becomes simpler – no points flying around, cache state is localized to a single cmap and although total LOC remains same, ~250 lines of new one are cmap data structure and writoutQueue comments. Most of previously existed modules enjoy significant reduction in LOC due to simplified logic.

redbaron:
Removed dependency between cache and persister/whisper modules

redbaron:
@lomik ,could you please have a look at this PR at least on patch-by-patch basis? 

It's been rebased twice privately which delayed it by a week and I doubt I can continue doing so for long. I'd have to keep it  forked otherwise and unfortunately won't be able to submit improvements upstream as that would require preparing 2 version of patches.

lomik:
In general, the changes seem to me interesting. But some things I would have implemented differently. When I get time I will start gradually transferring changes using your code chunks. If you do not mind of course.

redbaron:
I'll rebase and squash

redbaron:
Rebased & squashed. Also fixed all tests on per commit, so each one of them here passes all the tests.

I suggest put it into a branch and modify it to your taste if proposed changes are not far off your vision, it was last time I had a chance to dedicate time to brush up this patchset :(

lomik:
The basic ideas of the new architecture implemented in master

PR #111, opened 2016-09-08
Resolves #44 unstable throttle tests

### LINKED ISSUE #44: FAIL: TestThrottleChan on some machines (i.e. in virtualbox)
When i try make deb package over 

<pre>
git checkout
git submodules init
git submodules update
make clean && make && make deb
</pre>

i have following error on some machines

<pre>
 --- FAIL: TestThrottleChan (8.00 seconds)
     assertions.go:156:
     Location:       throttle_test.go:38
     Error:          Should be true
     Messages:       perSecond: 1000, bw: 822
</pre>

possible reason it happening when machine have too small CPU resources
do you have .deb PPA ? or maybe i can skip test? or maybe need increase minimum for throttling?

drawks:
- added an initial "unthrottled" run to determine max bandwidth
- skip tests that are greater than 90% of max
- added more values to test matrix

### DIFF
--- a/persister/throttle_test.go
+++ b/persister/throttle_test.go
@@ -2,7 +2,6 @@ package persister
 
 import (
 	"fmt"
-	"os"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


lomik:
Did not help

```
Maximum bandwidth is 482258. Skipping 800000 and higher.
--- FAIL: TestThrottleChan (9.00s)
    Location:   throttle_test.go:48
    Error:      Should be true
    Messages:   perSecond: 400000, bw: 317161

```

drawks:
Oh, I misunderstood the problem here. I thought we were solving for a limit which is set higher than the possible throughput on the test hardware.

What is actually happening here is that Ticker is imprecise at high resolutions.
This code should be converted to use something like https://godoc.org/golang.org/x/time/rate if you want better throttle precision for rate limiting than what you're getting.

I can't find where the reference, but I've seen mention of Ticker being only stable at rates of a few thousand/sec

PR #113, opened 2016-09-21
Fix deadlock between collector and app.ReloadConfig

redbaron:


lomik:
missed app.Unlock in 

```
if err = app.configure(); err != nil {
        return err
    }
```

redbaron:
you are right. decided to leave defer in there to protect from errors. worst thing if `collector.Stop()`  panics, then defer will try to unlock already  unlocked lock, generating another panic :)

lomik:
Refactored Collector without app.Lock - https://github.com/lomik/go-carbon/commit/8a780f715bba2f713a7dd75ac264ae8f8f974c06
(untested)

redbaron:
if all modules are safe to .Stat when they are stopped (or while they are being stopped), it looks like it is going to work

redbaron:
there is now an implicit not very obvious hard requirement to call NewCollector(app) on any change of app modules, it wasn't the case before though :(

lomik:
Added warning :) https://github.com/lomik/go-carbon/commit/7bd9da4bd74e8fb56345c990a9c14cd81f417d23

PR #116, opened 2016-10-14
Logrotate

jriguera:
Logrotate configuration for DEB and RPM packages


lomik:
I do not think that logrotate configuration must come with the rpm. Sorry.

Issue #128, opened 2016-11-24
[Packaging] RPM update to support systemd-based OS


mgrzybek:
Hi,

I have updated some files to create a RPM package that deals with systemd:
* conf-examples/carbon.conf: path updates for logs, data and schema
* deploy/rpm: added logrotate + service file

In order to create a clean PR, I have some questions:

1. Do you want a spec file per distribution release?
   * go-carbon.spec.centos, go-carbon.spec.suse...

2. Do you want a generic logrotate file?
   * buildrpm.sh could create it on the fly depending on carbon.conf and the target OS

Mathieu

lomik:
Hi,

I already have refactoring plan. All paths will be changed in release 0.10 (not next 0.9 release).

0.10 roadmap:
* Replace logging library from `logrus` to `uber-go/zap`. Remove reopen by HUP and inotify. Add periodic logfile changed check
* Move /usr/local/bin/go-carbon to /usr/bin/go-carbon
* Set config paths to `/etc/go-carbon/go-carbon.conf`, `/etc/go-carbon/storage-schemas.conf`, `/etc/go-carbon/storage-aggregation.conf`
* Add systemd service file

I do not like to include log rotation logic into package. But I still think about it

mgrzybek:
I have pushed some stuff in my fork. I can create a PR if you want. Fell free to comment: https://github.com/mgrzybek/go-carbon/commit/d0641aeb5f367b8b3c4289c5a3f24cd6e877e025

bzed:
I'll happily handle the systemd service integration in the debian packaging, but please merge the changes from @mgrzybek first to avoid double work.

blysik:
Any progress on this?

lomik:
systemd service has been added in 0.10.0 release

PR #132, opened 2016-12-04
Packaging - systemd-aware RPM spec file

mgrzybek:
According to #128
This PR adds a systemd service file, the required configuration files and renames the "go-carbon.centos.spec" to "go-carbon.spec"


lomik:
Thanks! I will merge this after 0.9 release

xneo64:
@lomik why is this pushed to 0.10? 
With the USR2 restart feature present in 0.9, I think this PR is essential for the 0.9 release

blysik:
I'm looking to deploy go-carbon onto RHEL7 servers.  Any idea when this will be merged?  Thanks!

AlexAkulov:
- Do you really need `_` prefix in `_graphite` username?
- You should include `network-online.target` in `After` option. Otherwise go-carbon could run before network.
- My config file also include this settings, they seem to be useful
```
TimeoutStopSec=600 # Default value in CentOS is 90s. Because dump saving may take longer, you should increase this number.
LimitNOFILE=55555 # Default limit is 1024. If you don't use relay in front of go-carbon, you might hit this limit.
LimitMEMLOCK=infinity # Cache size may be huge
```

mgrzybek:
The Debian graphite package creates a user called _graphite. That's why I used it there.
I have added your settings.

Moreover I think the configuration directory should be `/etc/carbon` instead of `/etc/go-carbon` to be compliant with graphite-project. Any idea ?

lomik:
> The Debian graphite package creates a user called _graphite. That's why I used it there.
I have added your settings.

Creating a user when installing the package is one of the issues on which I have not formed a final opinion.

> Moreover I think the configuration directory should be /etc/carbon instead of /etc/go-carbon to be compliant with graphite-project. Any idea ?

Projects are independent and have incompatible configs. Moreover, you can run them simultaneously.

hnakamur:
I'm using go-carbon on CentOS 7. This RPM package would be great!
Thanks for this pull request!

> Creating a user when installing the package is one of the issues on which I have not formed a final opinion.

I think it's fairly common for packages to create a user when installed
so many people would expect a user is created.
However I disagree `_` prefix in user names for rpm packages since it's uncommon.
In CentOS 7, use names without `_` prefix such as `nginx` or `apache` are used.

hnakamur:
Thanks for changing username from _graphite to carbon for RPM!

deploy/go-carbon.spec
```diff
@@ -0,0 +1,78 @@
...
+Group:		Development/Tools
+License:	MIT License
+URL:		https://github.com/lomik/go-carbon
+Source0:	%{name}.tar.gz
+BuildRoot:      %{_tmppath}/%{name}-%{version}
+
+%if systemd_requires
+%systemd_requires
+%else
+Requires:  /etc/init.d/functions
```
ehlerst: Can we change this to "initscripts" this breaks yum dependencies  in amazon linux.

```
Error: Package: go-carbon-0.9.1-1.el6.x86_64 (@timsrepo-rpm-repo)
           Requires: /etc/init.d/functions
           Removing: initscripts-9.03.49-1.34.amzn1.x86_64 (installed)
               Not found
           Updated By: initscripts-9.03.49-1.35.amzn1.x86_64 (amzn-updates)
               Not found
yum whatprovides /etc/init.d/functions
Loaded plugins: priorities, s3iam, update-motd, upgrade-helper
initscripts-9.03.49-1.34.amzn1.x86_64 : The inittab file and the /etc/init.d scripts
Repo        : installed
Matched from:
Filename    : /etc/init.d/functions
```

deploy/go-carbon.spec
```diff
@@ -0,0 +1,78 @@
...
+
+%{__install} -pD -m 755 %{name} %{buildroot}/%{_bindir}/%{name}
+%{__install} -pD -m 644 %{name}.conf %{buildroot}/%{_sysconfdir}/%{name}/%{name}.conf
+%{__install} -pD -m 644 storage-schemas.conf %{buildroot}/%{_sysconfdir}/%{name}/storage-schemas.conf
+%{__install} -pD -m 644 storage-aggregation.conf %{buildroot}/%{_sysconfdir}/%{name}/storage-aggregation.conf
+
+%if systemd_requires
+%{__install} -pD -m 644 %{name}.service %{buildroot}%{_unitdir}/%{name}.service
+%else
+%{__install} -pD -m 755 %{name}.init %{buildroot}/etc/init.d/%{name}
```
ehlerst: use the %{_initddir} macro here instead of /etc/init.d

lomik:
I think to keep demonization and user configuration in go-carbon. Otherwise systemd should create and chown log, data and other directories (and user should support same settings in go-carbon config and systemd service file).

I'm testing this systemd service file
```
[Unit]
Description=Golang implementation of Graphite/Carbon server.
Documentation=https://github.com/lomik/go-carbon
After=network.target

[Service]
Type=forking
ExecStart=/usr/bin/go-carbon -config /etc/go-carbon/go-carbon.conf -pidfile /var/run/go-carbon.pid -daemon
ExecReload=/bin/kill -HUP $MAINPID
KillSignal=USR2
Restart=on-failure

TimeoutStopSec=600
LimitNOFILE=55555
LimitMEMLOCK=infinity

[Install]
WantedBy=multi-user.target
```

ehlerst:
@lomik if you do -daemon the pid returns an exit 0 and systemd gets confused killing the pid and trying to restart it. I had tried this before submitting my merge for the same, i think they expect everything in the foreground. 

Edit: its from my systemd merge but its still the same problem.
ExecStart=/usr/local/bin/go-carbon -config /usr/local/etc/go-carbon.conf -pidfile /var/run/go-carbon.pid -daemon

 ```
go-carbon.service - go-carbon a golang implementation of Graphite/Carbon server
   Loaded: loaded (/usr/lib/systemd/system/go-carbon.service; disabled; vendor preset: disabled)
   Active: failed (Result: start-limit) since Tue 2017-04-04 17:26:44 UTC; 2s ago
  Process: 22041 ExecStart=/usr/local/bin/go-carbon -config /usr/local/etc/go-carbon.conf -pidfile /var/run/go-carbon.pid -daemon (code=exited, status=0/SUCCESS)
 Main PID: 22041 (code=exited, status=0/SUCCESS)

Apr 04 17:26:44 : go-carbon.service holdoff time over, scheduling restart.
Apr 04 17:26:44 : start request repeated too quickly for go-carbon.service
Apr 04 17:26:44 : Failed to start go-carbon a golang implementation of Graphite/Carbon server.
Apr 04 17:26:44 : Unit go-carbon.service entered failed state.
Apr 04 17:26:44 : go-carbon.service failed.
```

Without -daemon

```
 systemctl start go-carbon
 systemctl status go-carbon
● go-carbon.service - go-carbon a golang implementation of Graphite/Carbon server
   Loaded: loaded (/usr/lib/systemd/system/go-carbon.service; disabled; vendor preset: disabled)
   Active: active (running) since Tue 2017-04-04 17:29:16 UTC; 5s ago
 Main PID: 22116 (go-carbon)
   CGroup: /system.slice/go-carbon.service
           └─22116 /usr/local/bin/go-carbon -config /usr/local/etc/go-carbon.conf

Apr 04 17:29:16  systemd[1]: Started go-carbon a golang implementation of Graphite/Carbon server.
Apr 04 17:29:16  systemd[1]: Starting go-carbon a golang implementation of Graphite/Carbon server...
Apr 04 17:29:18 systemd[1]: Started go-carbon a golang implementation of Graphite/Carbon server.
```

lomik:
@ehlerst,
Do you set Type=forking?

tehlers320:
@lomik nope! i retract my comments. it works with Type=forking.

deploy/go-carbon.spec
mgrzybek: Change made.

PR #214, opened 2017-12-13
FreeBSD compatibility

dch:
> $ go version
> go version go1.9.2 freebsd/amd64
> $ uname -a
> FreeBSD wintermute.skunkwerks.at FreeBSD 12.0-CURRENT 
> $ git --version
> git version 2.15.1

I needed 2 changes to compile successfully here:

- trailing / on the .gitmodules causes checkout to fail, possibly only on certain versions of git
- `Bavail` can be negative on some systems and therefore has a `int64` type at least here

This is my first go code ever so I hope it's OK :-)

### DIFF
--- a/carbonserver/carbonserver.go
+++ b/carbonserver/carbonserver.go
@@ -472,7 +472,10 @@ func (listener *CarbonserverListener) updateFileList(dir string) {
 		return
 	}
 
-	freeSpace := stat.Bavail * uint64(stat.Bsize)
+	var freeSpace uint64
+	if stat.Bavail >= 0 {
+		freeSpace = uint64(stat.Bavail) * uint64(stat.Bsize)
+	}
 	totalSpace := stat.Blocks * uint64(stat.Bsize)
 
 	fileScanRuntime := time.Since(t0)
--- a/.gitmodules
+++ b/.gitmodules
@@ -135,7 +135,7 @@
 	url = https://github.com/rcrowley/go-metrics
 [submodule "vendor/github.com/davecgh/go-spew"]
 	path = vendor/github.com/davecgh/go-spew
-	url = https://github.com/davecgh/go-spew/
+	url = https://github.com/davecgh/go-spew
 [submodule "vendor/github.com/eapache/go-resiliency"]
 	path = vendor/github.com/eapache/go-resiliency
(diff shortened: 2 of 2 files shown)


carbonserver/carbonserver.go
```diff
@@ -472,7 +472,12 @@ func (listener *CarbonserverListener) updateFileList(dir string) {
 		return
 	}
 
-	freeSpace := stat.Bavail * uint64(stat.Bsize)
+	var freeSpace uint64
+	if stat.Bavail >= 0 {
+		freeSpace = uint64(stat.Bavail) * uint64(stat.Bsize)
+	} else {
+		freeSpace = 0
```
theckman: `var freeSpace uint64` means the value of `freeSpace` is `0`. You should remove the `else` block entirely as it's not needed.

theckman (changes requested):

carbonserver/carbonserver.go
dch: ^5 you're right, thanks!

carbonserver/carbonserver.go
```diff
@@ -472,7 +472,10 @@ func (listener *CarbonserverListener) updateFileList(dir string) {
 		return
 	}
 
-	freeSpace := stat.Bavail * uint64(stat.Bsize)
+	var freeSpace uint64
+	if stat.Bavail >= 0 {
```
theckman, line 476: Was looking at this again, and think this may be better as `stat.Bavail > 0`. If the value is zero, `n * 0` is always zero. So, `var freeSpace uint64` takes care of that.

theckman (changes requested):

Civil:
I wonder what happened so Bavail was <0 ?

dch:
@civil the main issue is that the types don't match so the code doesn't even compile. I wonder why nobody else already ran into that. The *reason* the types don't match is that some unix filesystems (linux ext4, for one) can have negative blocks free, if space is reserved via [tune2fs](https://linux.die.net/man/8/tune2fs) , and therefore `Bavail` needs to take that into account by using `int64` type.

e.g.:

```
$ sudo tune2fs -l /dev/sda1 | grep -i 'Reserved block count'
Reserved block count:     36743890
```

PR #220, opened 2018-01-17
Add compression support for receiver/tcp

bom-d-van:
cc @Civil

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -169,3 +169,6 @@
 [submodule "vendor/github.com/klauspost/compress"]
 	path = vendor/github.com/klauspost/compress
 	url = https://github.com/klauspost/compress.git
+[submodule "vendor/github.com/klauspost/cpuid"]
+	path = vendor/github.com/klauspost/cpuid
+	url = https://github.com/klauspost/cpuid.git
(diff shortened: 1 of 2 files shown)


obfuscurity:
This option is exposed with `-config-print-default` but doesn't appear to be documented anywhere. I assume the supported values are one of `gzip, zip, zlib, flate` (taken from [here](https://github.com/klauspost/compress#usage))?

deniszh:
@obfuscurity : according to https://github.com/lomik/go-carbon/pull/220/files only `gzip` and `snappy` is supported. Also, please note that is only merged in master and not included in any release yet.

obfuscurity:
Ah right, I missed that. Thanks @deniszh.

obfuscurity:
@deniszh Actually, it would appear that compression support is in v0.13.0.

deniszh:
Ah, right. I misread stuff, sorry.

obfuscurity:
No problem. 😄 ❤️

deniszh:
Not sure which relay supports compression, though. Carbon-c-relay should
support it, but not sure if it’s compatible....

PR #221, opened 2018-01-24
add support for Google PubSub

joemiller:
This PR adds support for ingesting metrics from google pubsub.

I consider it MVP at this point. 

Some notes on possible future changes:

* We may want to expose some subscriber setting tweaks via the config file (https://godoc.org/cloud.google.com/go/pubsub#ReceiveSettings)
* There may be some desire to have (possibly as an option) go-carbon create and delete subscriptions on its own. Currently it requires a subscription to be setup ahead of time which is a pattern that works for us and supports restarting go-carbon and minimizing dropped metrics.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -172,3 +172,21 @@
 [submodule "vendor/github.com/klauspost/cpuid"]
 	path = vendor/github.com/klauspost/cpuid
 	url = https://github.com/klauspost/cpuid.git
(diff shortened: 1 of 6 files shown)


Civil:
I think it might be a good idea to drop Go 1.8.x support in travis.

lomik:
wow, thank you!

PR #222, opened 2018-01-29
google pubsub gzip compression, and pubsub.ReceiverSettings

joemiller:
This PR adds two updates to the google pubsub support:

- https://github.com/lomik/go-carbon/commit/c4ea9a88f63c248281962d4fb40f566d661262bd Exposes the pubsub libraries [ReceiverSettings](https://godoc.org/cloud.google.com/go/pubsub#ReceiveSettings). I found this was necessary in benchmarking to achieve maximum throughput. Primarily, the num-go-routines at 4 instead of the default 1 was helpful in keeping go-carbon's buffer of messages full.

- https://github.com/lomik/go-carbon/commit/e9fda9603014383199896a7c8989beede4cb1d3c adds support for compressed pubsub message payloads. The codec is specified on the `codec` attribute of messages. Current values are either `none` (no compression) or `gzip`. Because google pubsub charges by the GB this makes a lot of sense. I also noted significant throughput improvements during load-testing with carbon-relay-ng(code from this PR: https://github.com/graphite-ng/carbon-relay-ng/pull/256)

### DIFF
--- a/README.md
+++ b/README.md
@@ -224,12 +224,21 @@ buffer-size = 0
 #
 # [receiver.pubsub]
 # # This receiver receives data from Google PubSub
-# # Authentication is managed through APPLICATION_DEFAULT_CREDENTIALS:
-# # - https://cloud.google.com/docs/authentication/production#providing_credentials_to_your_application
-# # Currently the subscription must exist before running go-carbon.
+# # - Authentication is managed through APPLICATION_DEFAULT_CREDENTIALS:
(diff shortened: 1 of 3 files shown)


lomik:
Please, edit config in deploy/go-carbon.conf and run python update_readme.py instead of direct editing in readme.md

joemiller:
Will do! I was not aware of that.

joemiller:
@lomik sorry about that. deploy/go-carbon.conf updated now.

PR #233, opened 2018-05-02
Fix info handler not returning numberOfPoints in Retentions

gksinghjsr:
Due to incomaptible type conversion between retention protocols v2 and v3, numberOfPoints was becoming 0. 
https://play.golang.org/p/zUJrIyB9792

Fix uses retention of first metric to construct v2 response, rather than trying to convert v3 to v2.

### DIFF
--- a/carbonserver/info.go
+++ b/carbonserver/info.go
@@ -8,8 +8,6 @@ import (
 	"strings"
 	"sync/atomic"
 	"time"
-	"unsafe"
-
 	"go.uber.org/zap"
 
 	"github.com/go-graphite/go-whisper"
@@ -19,9 +17,6 @@ import (
 	protov3 "github.com/go-graphite/protocol/carbonapi_v3_pb"
 )
 
-func retentionsV3toV2(ptr *[]protov3.Retention) *[]protov2.Retention {
-	return (*[]protov2.Retention)(unsafe.Pointer(ptr))
-}
 
 func (listener *CarbonserverListener) infoHandler(wr http.ResponseWriter, req *http.Request) {
 	// URL: /info/?target=the.metric.Name&format=json
@@ -88,7 +83,8 @@ func (listener *CarbonserverListener) infoHandler(wr http.ResponseWriter, req *h
 	)
 
 	response := protov3.MultiMetricsInfoResponse{}
-	for _, metric := range metrics {
+	var retentionsV2 []protov2.Retention
+	for i, metric := range metrics {
 		path := listener.whisperData + "/" + strings.Replace(metric, ".", "/", -1) + ".wsp"
 		w, err := whisper.Open(path)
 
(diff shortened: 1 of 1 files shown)


gksinghjsr:
Is there any reason why we were returning protoV2Format and protoV3Format as protoV2Format? If there are any concerns around this, then the pull request will need a change and go back to clubbing the two cases together, with the only change being related to RetentionV2 being used to construct the response.

gksinghjsr:
@Civil any ideas around grouping of protoV2Format and protoV3Format together?

carbonserver/info.go
```diff
@@ -135,12 +139,8 @@ func (listener *CarbonserverListener) infoHandler(wr http.ResponseWriter, req *h
...
-	case protoV2Format, protoV3Format:
-		if formatCode == protoV3Format {
-			contentType = httpHeaders.ContentTypeCarbonAPIv3PB
-		} else {
-			contentType = httpHeaders.ContentTypeCarbonAPIv2PB
-		}
+	case protoV2Format:
+		contentType = httpHeaders.ContentTypeCarbonAPIv2PB
 
 		var r protov3.MetricsInfoResponse
```
Civil, line 145: Shouldn't it be protov2.MetricsInfoResponse here?

Civil: Ah sorry, my fault, it's ok.

Civil:
@gksinghjsr what do you mean by "Is there any reason why we were returning protoV2Format and protoV3Format as protoV2Format" ?

For me current PR seems to be ok.

carbonserver/info.go
```diff
@@ -117,6 +113,14 @@ func (listener *CarbonserverListener) infoHandler(wr http.ResponseWriter, req *h
 				SecondsPerPoint: spp,
 				NumberOfPoints:  nop,
 			})
+			//only one metric is enough - first metric
```
Civil, line 116: ProtoV2 doens't support multiple metrics in response here.

I also think it makes more sense to move this sutff to `case protoV2Format`, just grab the first response from protoV3 structure and convert it to protoV2

Civil (changes requested):

gksinghjsr:
Whether metric is read from the form, or in case of protoV3, from body request, it will be an array, and it needs to be outside as even for protoV2, it is getting populated when the whisper files are being read. The protoV3 response is also the one being used as deafult to convert to json.

Civil:
I mean there is no need to populate protoV2 responses all the time. I think it's better to convert protoV3 to protoV2 on demand.
In that case we'll spend more time doing protov2 requests, but when people migrate to protov3 it won't do unnecessary work.

gksinghjsr:
@Civil Added a check to make sure that we update the retentions only in case the format is protoV2.

PR #234, opened 2018-05-15
Fix fetch response in prepareDataProto.

ibuclaw:
Looks like a logic regression bug.  Before it checked if there were no results after _all_ targets had been looked up.

https://github.com/lomik/go-carbon/blob/a19deb850326715e3f846de6acd83094f7030b8c/carbonserver/carbonserver.go#L1276-L1282

I need a couple days to verify this actually fixes it though. :-)

### DIFF
--- a/carbonserver/render.go
+++ b/carbonserver/render.go
@@ -303,33 +303,34 @@ func (listener *CarbonserverListener) prepareDataProto(format responseFormat, ta
 				continue
 			}
 			multiv2.Metrics = append(multiv2.Metrics, res.Metrics...)
-
-			if len(multiv2.Metrics) == 0 {
-				return fetchResponse{nil, contentType, 0, 0, 0, nil}, err
+		} else {
+			res, err := listener.fetchDataPB3(metric.Name, files, leafs, fromTime, untilTime)
+			if err != nil {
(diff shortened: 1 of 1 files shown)


Civil:
I think it's still good idea to merge this PR, but I'd like to understand if it actually affects the result?

ibuclaw:
@Civil what I saw was that given i.e: `/render/?target=host.one&target=host.two`.

Currently, if the first target returns no data, then `prepareDataProto` exits early, and ultimately returns nothing.  Swap the targets around and data is returned.

ibuclaw:
In the middle of re-reading this, yes this patch is definitely wrong.  The continue in the loop is fine, the check for `metricsFetched` is ok.  But the `+=` is not what I want, as that is an array that's appended to on each loop.

And I see that is also the case for `metrics`, `memoryUsed` and `valuesFetched` too.  All of that code should be hauled out of the range loop and done immediately after.

Civil:
Ah, right, thanks!

I'll wait for you to confirm that your patch works

Civil (approved):

ibuclaw:
@Civil - confirmed, this patch works.

PR #235, opened 2018-05-25
Enabling RealSize field in protov3.MetricDetails and switching Size a…

azhiltsov:
…nd RealSize

With introduction of RealSize it is possible to see both Size and RealSize of
sparse files, for examle in /metrics/details/?format=json.
Note: In case of full file allocation the RealSize can be bigger then the Size
because of the way filesystem allocates space on a disk.

linux stat utilities returns for test.wsp :
Size: 662440 Blocks: 16
/metrics/details will return:
Size: 662440 RealSize: 8192
where the RealSize is calculated as Blocks * 512


azhiltsov:
@Civil

Civil (approved):

PR #243, opened 2018-08-30
Fix hard limiter goroutine leak

gunnihinn:


gunnihinn:
Happily, we can confirm this works as expected, as yesterday we did *not* have a repeat of the outage that made us write these patches. :)

PR #251, opened 2018-09-12
carbonserver: support graphite-web's `REMOTE_STORE_USE_POST`

Quentin-M:
While `Content-Type: application/x-www-form-urlencoded` is supported, graphite-web sends POST requests with `Content-Type: multipart/form-data`.
Because we're explicitly calling `req.ParseForm()`, we are unable to process `multipart/form-data`.
Instead, we should remove `req.ParseForm()` entirely as calls to `FormValue()` automatically call `ParseMultipartForm()` / `ParseForm()` as necessary.

Ref: https://golang.org/pkg/net/http/#Request.FormValue
Fixes https://github.com/lomik/go-carbon/issues/249

### DIFF
--- a/carbonserver/render.go
+++ b/carbonserver/render.go
@@ -67,8 +67,6 @@ func stringToInt32(t string) (int32, error) {
 }
 
 func getFormat(req *http.Request) (responseFormat, error) {
-	req.ParseForm() // idempotent
-
 	format := req.FormValue("format")
 	if format == "" {
 		format = "json"
@@ -90,8 +88,6 @@ func getFormat(req *http.Request) (responseFormat, error) {
 }
 
 func getTargets(req *http.Request, format responseFormat) (map[timeRange][]target, error) {
-	req.ParseForm() // idempotent
-
 	targets := make(map[timeRange][]target)
 
 	switch format {
@@ -147,8 +143,6 @@ func (listener *CarbonserverListener) renderHandler(wr http.ResponseWriter, req 
 		zap.String("url", req.URL.RequestURI()),
 		zap.String("peer", req.RemoteAddr),
 	))
-
-	req.ParseForm()
 
 	format, err := getFormat(req)
 	if err != nil {
--- a/carbonserver/capability.go
+++ b/carbonserver/capability.go
@@ -29,7 +29,6 @@ func (listener *CarbonserverListener) capabilityHandler(wr http.ResponseWriter, 
 		zap.String("peer", req.RemoteAddr),
 	))
 
-	req.ParseForm()
 	format := req.FormValue("format")
 
 	accepts := req.Header["Accept"]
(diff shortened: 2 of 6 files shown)


Civil:
"To access multiple values of the same key, call ParseForm and then inspect Request.Form directly."

I think removing ParseForm breaks ability to pass multiple targets in a query. So this shouldn't be done in all the places. Maybe something a bit more clever and format-dependant at least in getTargets func.

I'm also not sure how behavior will change if you specify e.x. `&format=json&format=pickle`

carbonserver/render.go
```diff
@@ -90,8 +88,6 @@ func getFormat(req *http.Request) (responseFormat, error) {
 }
 
 func getTargets(req *http.Request, format responseFormat) (map[timeRange][]target, error) {
-	req.ParseForm() // idempotent
```
Civil, line 93: I think here we need to have more clever logic, because there might be multiple `target=` fields.

Civil (changes requested):

carbonserver/render.go
Quentin-M: It actually works just fine (see below). Our first call to `req.FormValue()` to get `from` and `until` will call `req.ParseMultipartForm()`, itself calling `req.ParseForm` as necessary. Therefore, `req.Form` is filled when we start iterating over `eq.Form["target"]`

## Content-Type: application/x-www-form-urlencoded

```
$ curl -vv -d "format=pickle;local=1;noCache=1;from=1536704479;until=1536715279;now=1536715279;target=carbon.agents.graphite-cache-0.cache.maxSize&target=carbon.agents.graphite-cache-1.cache.maxSize" "localhost:8080/render/"
```

```
Found target: carbon.agents.graphite-cache-0.cache.maxSize
Found target: carbon.agents.graphite-cache-1.cache.maxSize

[2018-09-12T17:25:37.544-0700] INFO [access] fetch served {"handler": "render", "url": "/render/", "peer": "[::1]:54108", "targets": ["carbon.agents.graphite-cache-0.cache.maxSize", "carbon.agents.graphite-cache-1.cache.maxSize"], "format": "pickle", "runtime_seconds": 0.000221681, "query_cache_enabled": true, "from_cache": false, "metrics_fetched": 0, "values_fetched": 0, "memory_used_bytes": 0, "http_code": 200}
```

## Content-Type: multipart/form-data

```
$ curl -vv -F "format=pickle" -F "local=1" -F "noCache=1" -F "from=1536704479" -F "until=1536715279" -F "now=1536715279" -F "target=carbon.agents.graphite-cache-0.cache.maxSize" -F "target=carbon.agents.graphite-cache-1.cache.maxSize" "localhost:8080/render/"
```

```
Found target: carbon.agents.graphite-cache-0.cache.maxSize
Found target: carbon.agents.graphite-cache-1.cache.maxSize

[2018-09-12T17:26:05.073-0700] INFO [access] fetch served {"handler": "render", "url": "/render/", "peer": "[::1]:54113", "targets": ["carbon.agents.graphite-cache-0.cache.maxSize", "carbon.agents.graphite-cache-1.cache.maxSize"], "format": "pickle", "runtime_seconds": 0.000284352, "query_cache_enabled": true, "from_cache": true, "metrics_fetched": 0, "values_fetched": 0, "memory_used_bytes": 0, "http_code": 200}
```

Civil (approved):

PR #254, opened 2018-09-13
Reply with 404 when nothing is found

gunnihinn:
These commits make go-carbon reply with a 404 HTTP status code when it can't find anything in its /render, /find and /info handlers instead of with a 200 code and an empty response. The exception is when we ask the /render handler for data in pickle format, in some attempt at compatibility with the Python stack.

This seems to be an OK use of the 404 code according to [RFC 7231](https://tools.ietf.org/html/rfc7231#section-6.5.4).

The motivation for this is that our request monitoring totally failed to pick up on a recent outage we had, where we managed to kill all our stores with a bad configuration change. Carbonzipper would get no data, and report empty responses with 200 codes to carbonapi. We'll make relevant patches to our end of the stack, but figured it would be fine to patch the entire pipeline.

If we're worried about this breaking backwards compatibility with something, we'd be happy to hide this behind a config option that's disabled by default.

### DIFF
--- a/carbonserver/find.go
+++ b/carbonserver/find.go
@@ -142,15 +142,25 @@ func (listener *CarbonserverListener) findHandler(wr http.ResponseWriter, req *h
 		response, err = listener.findMetrics(logger, t0, formatCode, query)
 	}
 
-	if response == nil {
+	if err != nil || response == nil {
+		var code int
+		var reason string
+		if _, ok := err.(errorNotFound); ok {
+			reason = "Not Found"
+			code = http.StatusNotFound
+		} else {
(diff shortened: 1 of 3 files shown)


Civil (approved):

Civil:
One thing that might be worth checking - how graphite-web will handle this, but I expect it to be ok to be honest.

gunnihinn:
Does graphite-web (only) speak pickle? If so, it _should_ not do anything different from before. Admittedly this was yesterday and I'm a bit tired now, but IIRC there is a case in /render where the pickle case still responds with a 200 and empty response.

PR #257, opened 2018-09-21
Add ability to re-publish tags at a specific update interval

### LINKED ISSUE #256: Tags should be re-published periodically
Howdy,

Just like carbon-cache does (see below), tags should not only be sent upon metric creation but re-published every certain number of updates, as the tag database may not be as reliable/long-lived as the carbon storage backend

## [carbon.conf](https://github.com/graphite-project/carbon/blob/master/conf/carbon.conf.example)
```
# Tag update interval, this specifies how frequently updates to existing series will trigger
# an update to the tag index, the default setting is once every 100 updates
# TAG_UPDATE_INTERVAL = 100

# Tag hash filenames, this specifies whether tagged metric filenames should use the hash of the metric name
# or a human-readable name, using hashed names avoids issues with path length when using a large number of tags
# TAG_HASH_FILENAMES = True

# Tag batch size, this specifies the maximum number of series to be sent to graphite-web in a single batch
# TAG_BATCH_SIZE = 100

# Tag queue size, this specifies the maximum number of series to be queued for sending to graphite-web
# There are separate queues for new series and for updates to existing series
# TAG_QUEUE_SIZE = 10000
```

Quentin-M:
Fixes https://github.com/lomik/go-carbon/issues/256

**TODO**: 
- Items in `updateCounters` must have a (fixed) TTL to keep memory usage somewhat under control (even though the `*uint64` map capacity will never shrink - we can still get the GC to sweep the extra `uint64`) when go-carbon runs for extended periods of time and when the metrics/tags turnover is high; it means we also want to (re-)publish the tags if it's not in the map yet, as updates may be slower than the TTL period and that therefore we would never re-publish 
- Run `gofmt`

### DIFF
--- a/tags/tags.go
+++ b/tags/tags.go
@@ -5,6 +5,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"net/url"
+	"sync/atomic"
 	"time"
 
 	"go.uber.org/zap"
@@ -14,23 +15,27 @@ import (
 )
 
 type Options struct {
-	LocalPath      string
-	TagDB          string
-	TagDBTimeout   time.Duration
-	TagDBChunkSize int
+	LocalPath           string
+	TagDB               string
+	TagDBTimeout        time.Duration
+	TagDBChunkSize      int
+	TagDBUpdateInterval uint64
 }
 
 type Tags struct {
-	q      *Queue
-	qErr   error // queue initialization error
-	logger *zap.Logger
+	q             *Queue
+	qErr          error // queue initialization error
+	logger        *zap.Logger
+	options       *Options
+	updateCounter uint64
 }
 
 func New(options *Options) *Tags {
 	var send func([]string) error
 
 	t := &Tags{
-		logger: zapwriter.Logger("tags"),
+		logger:  zapwriter.Logger("tags"),
+		options: options,
 	}
 
 	u, urlErr := url.Parse(options.TagDB)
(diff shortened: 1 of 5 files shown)


lomik:
Carbon-cache doesn't store counter for each metric

https://github.com/graphite-project/carbon/blob/b8fd836488ff92da25a2db149d671a669d2be004/lib/carbon/writer.py#L68

Quentin-M:
Sounds simpler, and eventually consistent as long as the interval is scaled appropriately to the load. That’s what I initially had in my PR before I change it. Their initial implementation was simply using a random number to determine the likelihood of updating the tags.

I also notice that they use two different queues for creates and updates so that creates are prioritized and less likely to be rejected by the queue being full. Do we want to do this here?

Quentin-M:
Updated the PR with a single update counter!

tags/tags.go
```diff
@@ -69,11 +79,16 @@ func (t *Tags) Stop() {
...
-	} else {
+func (t *Tags) Add(value string, now bool) {
+	if t.q == nil {
 		t.logger.Error("queue database not initialized", zap.Error(t.qErr))
+		return
+	}
+	if !now {
+		atomic.AddUint64(&t.updateCounter, 1)
+	}
+	if atomic.CompareAndSwapUint64(&t.updateCounter, t.options.TagDBUpdateInterval, 0) || now {
```
lomik: You can write this as
```
if now || ( atomic.AddUint64(&t.updateCounter, 1) % t.options.TagDBUpdateInterval == 0 ) {
t.q.Add(value)
}
```

PR #266, opened 2018-12-20
Remove metrics when file name too long from being stuck in loop forever,

gksinghjsr:
and running into same error again and again.


lomik:
please use a safer variant:
```
if pathErr, isPathErr := err.(*os.PathError); isPathErr && pathErr.Err == syscall.ENAMETOOLONG {
    p.pop(metric)
}
```

PR #268, opened 2019-01-17
Fix up the /find API for use with pickles

jjneely:
metric_path rather than path
isLeaf rather than is_leaf

### DIFF
--- a/carbonserver/find.go
+++ b/carbonserver/find.go
@@ -344,8 +344,8 @@ func (listener *CarbonserverListener) findMetrics(logger *zap.Logger, t0 time.Ti
 				metricsCount++
 			}
 			m = make(map[string]interface{})
-			m["path"] = p
-			m["is_leaf"] = glob.Leafs[i]
+			m["metric_path"] = p
+			m["isLeaf"] = glob.Leafs[i]
 
 			metrics = append(metrics, m)
 		}


jjneely:
It seems odd that I'm the first to find this.  Which makes me question how many others are using graphite-web to query carbonserver directly?

Looks like the recommended path is to continue to use a local graphite-web and query the carbonlink protocol from go-carbon.

azhiltsov:
originally carbonserver (https://github.com/grobian/carbonserver 
) was a standalone binary and was supposed to run aside of carbon-cache.
I joined the team in 2016 and we did not use a pickle already. 
Graphite-web is supposed to talk pickle to carbonzipper (part of carbonapi currently https://github.com/bookingcom/carbonapi) which will use protobuf in its turn. 

So the full-blown setup will have:
go-carbon
|
carbonzipper - graphite-web
|
 \\_carbonapi

I doubt that anyone is using carbonserver with graphite-web

lomik:
Current version of graphite-web supports both "is_leaf" and "isLeaf": https://github.com/graphite-project/graphite-web/blob/master/webapp/graphite/finders/remote.py#L121
"is_leaf" is used for compatibility with old graphite-web versions (0.9.x)

lomik:
I deleted my previous comment because I was wrong. Current version of carbonserver compatible only with new graphite-web. After this fix it should be compatible with graphite-web 0.9.x too

@Civil , how about to merge this?

Issue #263, opened 2018-11-15
Instrument go-carbon with Prometheus


gunnihinn:
Apart from being delicious, delicious betrayal, having Prometheus monitor go-carbon is useful when running the system at scale to avoid a "who watches the watchmen" situation.

I have some work towards doing this in [a branch](https://github.com/gunnihinn/go-carbon/tree/gmagnusson/prometheus). I have basic system and Go runtime metrics already, along with some instrumentation of carbonserver.

This issue is here so that if people have strong feelings about doing this, we can talk about them now before we invest more time in this line of patches.

lomik:
I like the idea. It would also be nice to add opentracing support to the carbonserver (and to the carbonapi).

gunnihinn:
> It would also be nice to add opentracing support to the carbonserver (and to the carbonapi).

Sure. I've never actually worked on a system with tracing support, so I'd need some time to tinker with a toy example before doing the work, and I'm not sure when I'll have the cycles to do that. If anyone else feels inspired to add this support in the meantime, that'd be great.

obfuscurity:
Apologies for piggybacking on this issue, it's unclear to me if this is already resolved given that 0.14.0 adds support for the Prometheus exporter. Anyhoo...

Are there any plans to expose all of the existing go-carbon metrics (`carbon.agents.*`) via the exporter? Or is the thinking that anyone that would be using Prometheus and go-carbon in the same company is already using Grafana and can just pull from both data sources?

azhiltsov:
@obfuscurity The main reason was explained by the topic starter, we did not want to monitor our graphite stack with graphite.
We did not set an ultimate goal to have a parity between Prometheus and graphite metrics. So coverage isn't full. Especially in those parts which we are not using.   
We are slowly adding those which are most important for us, but I doubt that we will cover everything. 

You are welcome to contribute, or may I suggest you to open a new issue with a list of metrics you are interested in. This might increase a chance that someone (who has a courage and some spare cycles) will implement it.

PR #269, opened 2019-01-18
Gmagnusson/prometheus

gksinghjsr:
Prometheus instrumentation for go-carbon so that we can monitor go-carbon with Prometheus rather than itself.


gksinghjsr:
#263

Civil:
I wonder if it make sense to only do this for Carbonserver, instead of making that change global for the whole go-carbon?

In that case you could avoid calling Prometheus from carbonserver at all and do that on metric exporter level.

However I like the refactoring of carbonserver, it's something that should be done for a very long time.

P.S. Also unit test segfaults, likely needs to be fixed as well.

gksinghjsr:
Moving this thread to #271 and closing this one, as I did not have commit rights in this repo.

PR #287, opened 2019-04-20
Support Compressed Whisper Format

bom-d-van:
Hi, this is a companying PR for supporting compressed whisper format (cwhisper) in go-whisper: https://github.com/go-graphite/go-whisper/pull/1, more details about compressed whisper format could be found in that PR.

One thing to note is that the changes in go-whisper is made to be backward compatible so there is no interruptions for existing whisper files.

This PR is set to be draft so that after the changes is accepted in go-whisper, I will update the dependency.

Looking forward to have this changes accepted and all feedback is welcome. :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,6 +108,8 @@ enabled = true
 # Use hashed filenames for tagged metrics instead of human readable
 # https://github.com/lomik/go-carbon/pull/225
 hash-filenames = true
+# specify to enable/disable compressed format
+compressed = false
 
 [cache]
 # Limit of in-memory stored points (not metrics)
@@ -416,6 +418,7 @@ With settings above applied, best write-strategy to use is "noop"
 
 ## Changelog
 ##### master
+* Added new options and upgraded go-whisper library to have compressed format (cwhisper) support
 
 ##### version 0.14.0
 * Accept UDP messages in plain protocol without trailing newline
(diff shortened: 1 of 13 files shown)


azhiltsov:
It would be nice to have the example config in deploy/go-carbon.conf and README.md changes in this PR

bom-d-van:
HI @lomik , I have marked this PR ready to review now. Could you take a look when you are convenient and let me know if things looks alright to you?

azhiltsov (commented):
@bom-d-van
Shouldn't we recommend to use locking by default with compressed format?

lomik:
I do not see any obvious problems.

README.md
```diff
@@ -108,7 +108,7 @@ enabled = true
 # Use hashed filenames for tagged metrics instead of human readable
 # https://github.com/lomik/go-carbon/pull/225
 hash-filenames = true
-# specify to enable/disable compressed format
+# specify to enable/disable compressed format. IMPORTANT: Only one process/thread could write to compressed whisper files at a time, especially when you are rebalancing graphite clusters (with buckytools, for example), flock needs to be enabled both in go-carbon and your tooling.
```
bom-d-van, line 111: @azhiltsov you are right, do you think this is good enough for reminding users of this enabling flock?

bom-d-van:
> @bom-d-van
> Shouldn't we recommend to use locking by default with compressed format?

@azhiltsov I have updated the doc a bit, could you take a look?

README.md
azhiltsov: Looks good to me, but might be additionally to that the flock config parameter need to be explicitly set to true? 
what is your take on that, @lomik ?

bom-d-van: @lomik @azhiltsov I have make flock enabled by default in example and doc now.

bom-d-van:
@lomik @azhiltsov I think this is good enough to be merged! :D

PR #293, opened 2019-06-03
persister: handle empty whisper files caused by edge cases like serve reboot

bom-d-van:


azhiltsov (commented):
@bom-d-van might be worth to make it a config option? 
I guess there could be different opinions on how to deal with corrupted files. 
@lomik do you think it worth an effort?

bom-d-van:
@azhiltsov I have added a config for controlling these clean up logics and it's default to false.

@lomik could you review my changes? I think it's ready now.

PR #299, opened 2019-06-25
Populate metric details and access times only if stats are enabled

gksinghjsr:
are enabled. Otherwise, its just a waste of memory and computation.


azhiltsov:
Thanks, @gksinghjsr 

image
image

PR #303, opened 2019-07-08
Trie Index + DFA for supporting globbing/wildcard queries

bom-d-van:
DFA implementation inspired by rsc: https://swtch.com/~rsc/regexp/regexp1.html

An alternative metric indexing implementation to trigram index. This implementation is built to support 10+ millions metrics per go-carbon instance (which becomes feasible after cwhisper support).

Currently only benchmark against a subset of our production queries, but already shows good potential. 92% of the queries is 1.34 - 3300.71 times faster. And the (1.02 - 18.29 times) slower queries is because of using leading star in a query nodes, which is possible to be re-written to become faster than the original query in trigram.

But still, more tests and changes are on the way.

Linking relevant issues that might be able to be resolved by this new implementation: 
https://github.com/lomik/go-carbon/issues/202
https://github.com/lomik/go-carbon/issues/301


carbon/config.go
```diff
@@ -113,6 +113,7 @@ type carbonserverConfig struct {
 	FailOnMaxGlobs    bool      `toml:"fail-on-max-globs"`
 	MetricsAsCounters bool      `toml:"metrics-as-counters"`
 	TrigramIndex      bool      `toml:"trigram-index"`
+	TrieIndex         bool      `toml:"trie-index"`
```
azhiltsov: Since trie and trigram are mutually exclusive we need to prevent them from being enabled both at the same time.

bom-d-van: @azhiltsov you are fast as lightning. :D

When trie-index is enabled, trigram-index should be ignored in the code. But I need to double check.

dgryski:
Is there a point where the trigram index is better or is the trie index always superior?

azhiltsov:
@dgryski, 
@gksinghjsr wrote a performance test for the trigram lib. 
He probably could run it against trie to compare them in the same scenarios.

README.md
```diff
@@ -303,12 +305,27 @@ trigram-index = true
...
+trie-index = false
+trie-index-with-trigrams = false
+
 # Maximum amount of globs in a single metric in index
 # This value is used to speed-up /find requests with
 # a lot of globs, but will lead to increased memory consumption
 max-globs = 100
 # Fail if amount of globs more than max-globs
 fail-on-max-globs = false
+# Maximum metrics could be returned by glob/wildcard in find request
```
azhiltsov: What will be returned in the failure scenario here?

README.md
```diff
@@ -303,12 +305,27 @@ trigram-index = true
...
+max-metrics-globbed  = 30000
+# Maximum metrics could be returned in render request
```
azhiltsov: What will be returned in the failure scenario here?

README.md
```diff
@@ -303,12 +305,27 @@ trigram-index = true
...
 scan-frequency = "5m0s"
+# Control trie index
+#  This index is built as an alternative to trigram index, with shorter indexing
+#  time and less memory usage (around 2 - 5 times). For most of the queries, trie
+#  is faster than trigram. For queries with keyword wrap around by widcards (like
+#  ns1.ns2.*keywork*.metric), trigram indexi performs better. Trie index could
+#  be speeded up by enabling adding trigrams to trie, at the cost of memory usage (
+#  depends on the facotr, the memory usage of trie would be at least doubled).
+trie-index = false
+trie-index-with-trigrams = false
```
azhiltsov: Would it be possible to use 
trigram-index = true && trie-index = true 
in order to specify
trie-index-with-trigrams = true
? 
3 parameters in the config creates a complexity of understanding which combination of them is acceptable and which not.

carbonserver/carbonserver.go
```diff
@@ -699,52 +750,13 @@ func (listener *CarbonserverListener) expandGlobs(query string) ([]string, []boo
 	}
 	globs = append(globs, query)
 	// TODO(dgryski): move this loop into its own function + add tests
```
azhiltsov: will this comment be relevant after merge?

README.md
```diff
@@ -303,12 +305,27 @@ trigram-index = true
...
+
 # Maximum amount of globs in a single metric in index
 # This value is used to speed-up /find requests with
 # a lot of globs, but will lead to increased memory consumption
 max-globs = 100
```
azhiltsov, line 322: would this and corresponding code become redundant with introduction of max-metric-globbed and max-metrics-rendered?

README.md
bom-d-van: Now it just silently truncated the returning result to this upper limit for easy implementation. I could tweak it if you have a better idea. :D

README.md
bom-d-van: It's not a bad idea. Let me think about it and see if it's possible/good.

carbonserver/carbonserver.go
bom-d-van: Nice catch. :P Will see what I could do.

README.md
bom-d-van: max-globs is used to control expanding {}, take the example bellow:

```
abc.{1,2,3,4}.{x,y,z}
->
abc.1.x
abc.1.y
...
abc.4.z
```

So this flag max-globs controls how many metrics is expanded to individual metrics, before the actual index query works. Its a flag only used in trigram-index and some special queries of trie-index.

But max-metric-globbed and max-metrics-rendered are used in truncating matched results, it works after the index query. I introduce them to prevent reading too many files and returning too many results that might lead to killing go-carbon. 

They are not necessarily part of trie-index works. But when I was testing trie-index on production queries, I see that trie-index is able to return thousands of metrics on some queries while trigram-index just failed, so I think it might be necessary to  have these control in place for the rendering production test of trie-index.

I admit it's a bit confusing, but they serve a different purpose.

gaoxp:
hi，when i use trie index instead of trigram index i find a problem, when i find metrics like a.b.\*.\*.\*.c, goroutine which serve this request will panic,do you have same problem?

also i find when go-carbon run one day or two day, carbonserver accour too many close-wait tcp connection, i don't understand why so many close-wait connections.

azhiltsov:
> also i find when go-carbon run one day or two day, carbonserver accour too many close-wait tcp connection, i don't understand why so many close-wait connections.

Trie enabled version of go-carbon is not in production in the Booking currently. We are running 9b6b1333 (current master) and don;t observe this problem. Can you confirm that you observe the same problem with 9b6b1333 and if so - open a separate issue?

bom-d-van:
> hi，when i use trie index instead of trigram index i find a problem, when i find metrics like a.b.*.*.*.c, goroutine which serve this request will panic,do you have same problem?

@gaoxp Is it possible for you to share us the stack trace?

gaoxp:
> > hi，when i use trie index instead of trigram index i find a problem, when i find metrics like a.b._._.*.c, goroutine which serve this request will panic,do you have same problem?
> 
> @gaoxp Is it possible for you to share us the stack trace?



> > hi，when i use trie index instead of trigram index i find a problem, when i find metrics like a.b._._.*.c, goroutine which serve this request will panic,do you have same problem?
> 
> @gaoxp Is it possible for you to share us the stack trace?


I try to fix this problem, the stack trace may different from original code,but it is sure that use trie index will cause this panic problem.
STACK TRACE:
2019/08/14 14:00:43 http: panic serving 10.120.197.17:38442: runtime error: invalid memory a
ddress or nil pointer dereference
goroutine 569 [running]:
net/http.(*conn).serve.func1(0xc4202180a0)
	/home/yd/golang/src/net/http/server.go:1726 +0xd0
panic(0xbcc120, 0x122ef90)
	/home/yd/golang/src/runtime/panic.go:505 +0x229
github.com/lomik/go-carbon/carbonserver.(*CarbonserverListener).expandGlobsTrie(0xc4202daa00
, 0xc42a41c0c0, 0x32, 0x1, 0xc42046a1b0, 0x9, 0x9, 0xc42a41c0c0, 0x32, 0x0, ...)
	/home/gaoxiaopeng/go-carbon/_vendor/src/github.com/lomik/go-carbon/carbonserver/trie
.go:862 +0x217
github.com/lomik/go-carbon/carbonserver.(*CarbonserverListener).expandGlobs(0xc4202daa00, 0x
c42a41c0c0, 0x32, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
	/home/gaoxiaopeng/go-carbon/_vendor/src/github.com/lomik/go-carbon/carbonserver/carb
onserver.go:840 +0x33d
github.com/lomik/go-carbon/carbonserver.(*CarbonserverListener).findMetrics(0xc4202daa00, 0x
c420405ce0, 0xbf4d06c2f720586b, 0x1b1d3386e, 0x1246780, 0x2, 0xc420212a40, 0x1, 0x1, 0x0, ..
.)
	/home/gaoxiaopeng/go-carbon/_vendor/src/github.com/lomik/go-carbon/carbonserver/find
.go:213 +0x159
github.com/lomik/go-carbon/carbonserver.(*CarbonserverListener).findHandler(0xc4202daa00, 0x
d95460, 0xc42cd1c020, 0xc42020e300)
	/home/gaoxiaopeng/go-carbon/_vendor/src/github.com/lomik/go-carbon/carbonserver/find
.go:143 +0x2059
github.com/lomik/go-carbon/carbonserver.(*CarbonserverListener).(github.com/lomik/go-carbon/
carbonserver.findHandler)-fm(0xd95460, 0xc42cd1c020, 0xc42020e300)
	/home/gaoxiaopeng/go-carbon/_vendor/src/github.com/lomik/go-carbon/carbonserver/carb
onserver.go:1189 +0x48
net/http.HandlerFunc.ServeHTTP(0xc42040e040, 0xd95460, 0xc42cd1c020, 0xc42020e300)
	/home/yd/golang/src/net/http/server.go:1947 +0x44
github.com/lomik/go-carbon/carbonserver.TraceHandler.func1(0xd989a0, 0xc420286000, 0xc42020e
200)
	/home/gaoxiaopeng/go-carbon/_vendor/src/github.com/lomik/go-carbon/carbonserver/carb
onserver.go:170 +0x2c1
github.com/lomik/go-carbon/vendor/github.com/dgryski/httputil.TimeHandler.func1(0xd989a0, 0x
c420286000, 0xc42020e200)
	/home/gaoxiaopeng/go-carbon/_vendor/src/github.com/lomik/go-carbon/vendor/github.com
/dgryski/httputil/times.go:26 +0x7f
github.com/lomik/go-carbon/vendor/github.com/dgryski/httputil.TrackConnections.func1(0xd989a
0, 0xc420286000, 0xc42020e200)
	/home/gaoxiaopeng/go-carbon/_vendor/src/github.com/lomik/go-carbon/vendor/github.com
/dgryski/httputil/track.go:40 +0xea
net/http.HandlerFunc.ServeHTTP(0xc42040e060, 0xd989a0, 0xc420286000, 0xc42020e200)
	/home/yd/golang/src/net/http/server.go:1947 +0x44
net/http.(*ServeMux).ServeHTTP(0xc420406030, 0xd989a0, 0xc420286000, 0xc42020e200)
	/home/yd/golang/src/net/http/server.go:2337 +0x130
github.com/lomik/go-carbon/vendor/github.com/NYTimes/gziphandler.GzipHandlerWithOpts.func1.1
(0xd989a0, 0xc420286000, 0xc42020e200)
	/home/gaoxiaopeng/go-carbon/_vendor/src/github.com/lomik/go-carbon/vendor/github.com
/NYTimes/gziphandler/gzip.go:293 +0x263
net/http.HandlerFunc.ServeHTTP(0xc420406120, 0xd989a0, 0xc420286000, 0xc42020e200)
	/home/yd/golang/src/net/http/server.go:1947 +0x44
net/http.serverHandler.ServeHTTP(0xc420434000, 0xd989a0, 0xc420286000, 0xc42020e200)
	/home/yd/golang/src/net/http/server.go:2694 +0xbc
net/http.(*conn).serve(0xc4202180a0, 0xd997a0, 0xc4200d72c0)
	/home/yd/golang/src/net/http/server.go:1830 +0x651
created by net/http.(*Server).Serve
	/home/yd/golang/src/net/http/server.go:2795 +0x27b

gaoxp:
> > also i find when go-carbon run one day or two day, carbonserver accour too many close-wait tcp connection, i don't understand why so many close-wait connections.
> 
> Trie enabled version of go-carbon is not in production in the Booking currently. We are running [9b6b133](https://github.com/lomik/go-carbon/commit/9b6b133373a3721c2561cb471f2fad194aa8228e) (current master) and don;t observe this problem. Can you confirm that you observe the same problem with [9b6b133](https://github.com/lomik/go-carbon/commit/9b6b133373a3721c2561cb471f2fad194aa8228e) and if so - open a separate issue?



> > also i find when go-carbon run one day or two day, carbonserver accour too many close-wait tcp connection, i don't understand why so many close-wait connections.
> 
> Trie enabled version of go-carbon is not in production in the Booking currently. We are running [9b6b133](https://github.com/lomik/go-carbon/commit/9b6b133373a3721c2561cb471f2fad194aa8228e) (current master) and don;t observe this problem. Can you confirm that you observe the same problem with [9b6b133](https://github.com/lomik/go-carbon/commit/9b6b133373a3721c2561cb471f2fad194aa8228e) and if so - open a separate issue?

I don't  observe the same problem with [9b6b133],just for trie-index version.

bom-d-van:
@gaoxp I am unable to reproduce your problem. But from the stack trace, it looks like trieIdx isn't initialized.

gaoxp:
> ur pro



> @gaoxp I am unable to reproduce your problem. But from the stack trace, it looks like trieIdx isn't initialized.

ok, I use trigram index instead of trie index when i find a.b.\*.\*.\*.f,and now it runs normal,thank you

carbonserver/carbonserver.go
bom-d-van: Updated the todo. ;)

gaoxp:
hello @bom-d-van, can you add fallback to filesystem logic when trie index not finish initial，becase i find when trie index not initialled, carbonserver not work well, and it will take some time to complete trie index initial,during this time, i can't see any metric use grafana.

bom-d-van:
@gaoxp it's a valid point, I will add the relevant logics.

bom-d-van:
> hello @bom-d-van, can you add fallback to filesystem logic when trie index not finish initial，becase i find when trie index not initialled, carbonserver not work well, and it will take some time to complete trie index initial,during this time, i can't see any metric use grafana.

@gaoxp I have pushed a simple fix for it. Though this is a problem shared by both trigram and trie indexing.

bom-d-van:
> Is there a point where the trigram index is better or is the trie index always superior?

@dgryski  after some production testing and more understandings that I gained while working on this, my answer to this question boils down to it depends.

There are at least one types of queries that I know of trigram performs better, that is if there is keywords extracted from the query that is able to narrow down the result. (For example, abc.*keyword1.xxx.yyy.zzz, and there are a high numbers of metrics matching this pattern abc.*.xxx.yyy.zzz), because there are leading stars in the query, trie+dfa would have too go through every one of them. (thus we incorporated some levels of trigrams in trie nodes and performs some early checks to make it works a little bit better for trie+dfa, (however, I think it might not be a very good and generic solution))

However, there is some clear advantage on trie+dfa, which is less memory consumption (around 1/5 of trigram) and faster indexing time (also around 1/5 of trigram). And in the majorities of the queries that I found in our production environment, trie+dfa is a clear winner. (in our latest test on one of the cluster, trie+dfa's 99th percentile of request time is around 1/5 of trigram)

@azhiltsov is helping on testing it on more our prod clusters to have more understandings and feedback.

bom-d-van:
@azhiltsov @lomik @dgryski I think this PR is mature enough for a formal review now. There might be unknown bugs and more optimizations could be done, but we could address them when we run into them.

dgryski:
Does this need fuzz tests for the query engine?

bom-d-van:
@dgryski It's a nice idea. Given both metric paths and queries are user inputs, fuzz test might help!

PR #307, opened 2019-07-22
Implement context cancellations for find and find part of render

gksinghjsr:
This pull request aims to introduce context cancellations to go-carbon. Currently, cancellations are implemented in api and zipper but not in go-carbon. This will help avoid uneeded work(in case client is gone), or when the request is taking too much time, and reduce load on the server.

### DIFF
--- a/carbonserver/carbonserver.go
+++ b/carbonserver/carbonserver.go
@@ -103,6 +103,14 @@ type QueryItem struct {
 	QueryFinished chan struct{}
 }
 
+//type GlobResponse struct {
+type ExpandedGlobResponse struct {
+	Name  string
+	Files []string
+	Leafs []bool
+	Err   error
+}
+
 var TraceHeaders = map[string]string{
 	"X-CTX-CarbonAPI-UUID":    "carbonapi_uuid",
 	"X-CTX-CarbonZipper-UUID": "carbonzipper_uuid",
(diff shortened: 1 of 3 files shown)


azhiltsov:
I tested this by executing queries to go-carbon directly.
TODO render requests (they are not getting cancelled)
find queries are getting cancelled
```
curl -v -o /dev/null '127.0.0.1:8080/metrics/find/?query=prod.*.*.*.org.*.*.*.*.%7BRead%2CWrite%7DLatency.*.*.*'
Ctrl+C
```
In the logs:
```
{
  "level": "ERROR",
  "timestamp": "2019-07-23T16:47:28.483+0200",
  "logger": "access",
  "message": "find failed",
  "handler": "find",
  "url": "/metrics/find/?query=prod.*.*.*.org.*.*.*.*.%7BRead%2CWrite%7DLatency.*.*.*",
  "peer": "127.0.0.1:52880",
  "query": [
    "prod.*.*.*.org.*.*.*.*.{Read,Write}Latency.*.*.*"
  ],
  "format": "json",
  "runtime_seconds": 1.191762335,
  "reason": "Internal error while processing request",
  "error": "find failed, timeout",
  "http_code": 500
}
```
What I am missing here: 
carbonapi_uuid - to track where the request came from
error - please remove it in favor of 'message' field and clear wording 'cancelled' or 'timeout' there
reason - not needed, please remove
http_code - Replace with HTTP 499

azhiltsov (approved):

PR #310, opened 2019-09-25
Added Prometheus metric to count number of incoming datapoints in TCP receiver.

grzkv:
This metric is quite useful in following cases:

- checking if storages receive expected number datapoints, e.g. if there is no significant data loss
- to monitor the incoming traffic for load estimation and fluctuations

Currently, the implementation is only for TCP. Can be added for other protocols as well.

### DIFF
--- a/carbon/app.go
+++ b/carbon/app.go
@@ -369,6 +369,10 @@ func (app *App) Start() (err error) {
 			return
 		}
 
+		if conf.Prometheus.Enabled {
+			rcv.InitPrometheus(app.PromRegisterer)
+		}
+
 		app.Receivers = append(app.Receivers, &NamedReceiver{
 			Receiver: rcv,
 			Name:     "tcp",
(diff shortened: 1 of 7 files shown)


lomik:
It might be better to use a counter instead of a gauge

grzkv:
I agree that a counter would be better. But, I have reused the implementation that was already in place for graphite. I did not add any new metric calculation.

Moving to a gauge would include re-writing the graphite metric calculation as well to avoid duplicated implementation. I think it makes sense to do it separately since it looks to fall out of the scope of this pull-request. Current PR is about exposing the already existing metric via a new interface.

lomik:
You can reuse current code for Counter too
```
if rcv.promMetricsReceived != nil {
    rcv.promMetricsReceived.Add(float64(metricsReceived))
}
```

grzkv:
@lomik You're right. I did the change you pointed out. Please have a look.

PR #329, opened 2019-12-03
Implement concurrent filepath.Walk

deniszh:
just properly implemented @KingAir idea from https://github.com/lomik/go-carbon/pull/327
didn't test it properly, though

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -2,7 +2,6 @@
 
 
 [[projects]]
-  digest = "1:250b2cc0a8f0e3ac641ce2c166bec9b51ddbbfd1b39cc2dc815cfbb97b8a1d17"
   name = "cloud.google.com/go"
   packages = [
     "compute/metadata",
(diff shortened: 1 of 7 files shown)


azhiltsov:
@deniszh @KingAir 
What are the benefits of using cwalk? Can you share any potential or real test cases? 
Is it always superior? I would expect it compete for resource with actual write/render requests.

Another concern here is https://github.com/iafan/cwalk#differences-from-filepathwalk
@bom-d-van does trigram/trie indexing requires sorted input?

bom-d-van:
> does trigram/trie indexing requires sorted input?

@azhiltsov  It might triggered some edge case issues/bugs. But in theory, both solution should be able to work for unordered input.

----

```
err := cwalk.Walk("/path/to/dir", walkFunc)
```

@deniszh better to be able to turn off/on this feature with a flag? Also, if the walkfunc is called concurrently, then we have some race conditions in the code.

deniszh:
TBH I have no idea. Just saw @KingAir PR and saw him as struggling with vendoring, so, created that. If he have no interest - we can close it.

deniszh:
I think it's not revelant anymore, will close it.

PR #332, opened 2020-01-14
trie: inline trie index building in filepath.Walk for saving memory overhead

bom-d-van:
On a node containing 40+ millions of metrics, this could cut down the memory usage
of go-carbon by up to ~50% (60+GB -> 30+GB).

NOTE: indexing_runtime will be 0 now for trie index as it is mixed in file_scan_runtime.

Because the current implementation of trigram expects an array input, so only change trie for the time being.


azhiltsov (approved):
Thank you for contribution

PR #336, opened 2020-02-25
Ephemeral nodes for persister (WIP)

bom-d-van:
Early preview of my yet another idea for go-carbon/go-graphite.

Pods comes and goes quickly on kubernetes and this creates issue for monitoring it with Graphite. Long term metrics is hard to be accommodated because most of the pods don't live long enough, thus preallocation of big metric file becomes too expensive.

This merge request attempts to **mitigate** the issue by reusing old and relevant metric files belongs to the same deployment/application/service by reusing old metric files.

For example, if an application have 3 pods, and each pod generates some metrics for cpu/memory/requests.

with this ephemeral node configuration:

```
k8s.*.$ephemeral.*
```

existing files in data dir:

```
k8s/app1/pod-id-01/cpu.wsp --> old file
k8s/app1/pod-id-02/cpu.wsp --> current file
k8s/app1/pod-id-02/mem.wsp
k8s/app2/pod-id-02/cpu.wsp
```

When a new cpu metric for a new pod for app1 is produced, go-carbon would check if there is old metric file for the same metric. If there is, instead of creating a new metric file, it would pick it up and reuse it. Note: it won't reset the file, it would only rename it and keep the history.

So for the above example, when a new metics named k8s.app1.pod-id-03.cpu comes in, go-carbon would rename k8s/app1/pod-id-01/cpu.wsp to k8s/app1/pod-id-03/cpu.wsp

Pros:

* long term history for short-lived pod metrics
* less stress on disk usage

Cons:

* if rollouts are too frequent, shorter than recycle threshold, then new files would still be generated

Codes still incomplete and production validation still needed. Just preview.

### DIFF
--- a/carbon/app.go
+++ b/carbon/app.go
@@ -93,6 +93,13 @@ func (app *App) configure() error {
 		} else {
 			cfg.Whisper.Aggregation = persister.NewWhisperAggregation()
 		}
+
+		if cfg.Whisper.EphemeralFilename != "" {
+			cfg.Whisper.ephemeralFilter, err = persister.NewEphemeralFilter(app.Config.Whisper.DataDir, cfg.Whisper.EphemeralFilename, cfg.Whisper.RecycleThreshold)
+			if err != nil {
+				return err
+			}
+		}
 	}
 	if !(cfg.Cache.WriteStrategy == "max" ||
 		cfg.Cache.WriteStrategy == "sorted" ||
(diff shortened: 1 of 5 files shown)


bom-d-van:
Another important note for this feature is that relays program and buckytools would also need to gain the capability of sending metrics belongs to the same application on kubernetes to the same go-carbon instance to maximize its benefit.

A simple design would be matching metrics by the ephemeral config and rewrite $ephemeral nodes to `_` and then use it to calculate its jump hash for finding backends.

```
config: k8s.*.$ephemeral.*

changes:
k8s.app1.pod-id-02.cpu
->
k8s.app1._.cpu
```

This could be counted as a Con for this change.

---

kubernetes is referred here exclusively as a design target, but this feature could be used for other scenarios.

deniszh:
Is this PR still relevant, @bom-d-van ?

bom-d-van:
Hi @deniszh, this one we can close it for now. Decided to pull off this feature on the producer side.

deniszh:
Cool, thanks!

Issue #318, opened 2019-11-10
delay before metrics are available for serving by carbonserver


luke-heberling:
Using v0.14.0, it seems like filesystem metadata is used for communicating all metric names between carbonserver and the cache, so there's a long delay of approximately (scan-frequency + queueWriteOutTime) before new metrics are available.

I'd like to figure out what is the expected delay for new metrics showing up in carbonserver queries, but I wasn't able to find it in the documentation. I see there's been a fair amount of development since v0.14.0, in particular https://github.com/lomik/go-carbon/pull/303 just a few days ago. Can someone help to clarify the current expectation for how/when new metrics will be available for queries?

bom-d-van:
@luke-heberling if you are using trigram index or trie index, your estimation is basically correct: scan-frequency + queueWriteOutTime + index build time (if you have few million metrics per host, it could take a few minutes for trigram index, trie is bit faster. but full disk scan is still part of the major overhead). 

It's possible to make it faster for new metrics to become available, by updating the index whenever a new metric is created.

luke-heberling:
Thanks @bom-d-van! Looking at the idea of updating the index when there are new metrics created, I see that there is filesystem metadata (e.g. modified time) included which appears to be for the proto v3 api. I'm not sure how the api clients need to use that info, and it's not obvious to me what carbonserver should be returning there for cache-only data. I'd like to get a proposal together for implementing this, can I get some advice on how to handle that?

bom-d-van:
I'm not really familiar with the metadata part, but for the advice of updating index:

The basic idea would be having a buffered channel (preferred, so index update could be done without blocking creating new files) or a callback passing into package persister, so it could notify index on adding new metric. There are a few tricky parts: 

* handling concurrency. As now index is replaced as a whole, so we don't have to worry about race condition. But to support real time index update, we need take that into consideration;
* we now have two index solutions, but for simplicity, you can start with just trigram or trie;
* have a config to disable/enable realtime index update;
* full disk scan still need to be kept as metrics could be removed by other tools or manually.

More problems or better solutions might be found while you are working on it, so these are just my simple suggestions. :)

luke-heberling:
[...]

PR #338, opened 2020-02-28
 cache-scan tunable to scan and populate cache-only metrics in the index

saipraneethyss:
This is related to #318 

With this implementation, the index gets updated with new metrics even if there is no whisper file created on disk. This significantly reduces the delay in making the new metrics available for queries.

This change provides a new config-parameter `cache-scan`. When enabled -
- the index (trie/trigram) is updated/populated with metrics, including the ones without wsp files on disk 
- carbonserver can execute and send responses for queries (in first retention rate) on cache-only metrics once the index is updated.
- it takes *at most* the file-scan-frequency duration for the new metrics to show up.

### DIFF
--- a/README.md
+++ b/README.md
@@ -318,8 +318,8 @@ trie-index = false
 
 # This provides the ability to query for new metrics without any wsp files
 # i.e query for metrics present only in cache. Does a cache-scan and
-# populates index with metrics with or without corresponding wsp files.
-# Disabled by default.
+# populates index with metrics with or without corresponding wsp files,
+# but will lead to increased memory consumption. Disabled by default.
 cache-scan = false
 
 # Maximum amount of globs in a single metric in index
--- a/README.md
+++ b/README.md
@@ -316,6 +316,12 @@ scan-frequency = "5m0s"
 #  memory usage (by setting both trie-index and trigram-index to true).
 trie-index = false
 
+# This provides the ability to query for new metrics without any wsp files
+# i.e query for metrics present only in cache. Does a cache-scan and
+# populates index with metrics with or without corresponding wsp files.
+# Disabled by default.
+cache-scan = false
+
 # Maximum amount of globs in a single metric in index
 # This value is used to speed-up /find requests with
 # a lot of globs, but will lead to increased memory consumption
--- a/cache/cache.go
+++ b/cache/cache.go
@@ -63,6 +63,7 @@ type Shard struct {
 	items            map[string]*points.Points
 	notConfirmed     []*points.Points // linear search for value/slot
 	notConfirmedUsed int              // search value in notConfirmed[:notConfirmedUsed]
+	adds             map[string]struct{} // map to maintain all the new metric names
 }
 
 // Creates a new cache instance
@@ -76,6 +77,7 @@ func New() *Cache {
 		c.data[i] = &Shard{
 			items:        make(map[string]*points.Points),
 			notConfirmed: make([]*points.Points, 4),
+			adds:         make(map[string]struct{}),
 		}
 	}
 
@@ -275,6 +277,7 @@ func (c *Cache) Add(p *points.Points) {
 		values.Data = append(values.Data, p.Data...)
 	} else {
 		shard.items[p.Metric] = p
+		shard.adds[p.Metric] = struct{}{}
 	}
 	shard.Unlock()
 
@@ -323,4 +326,21 @@ func (c *Cache) PopNotConfirmed(key string) (p *points.Points, exists bool) {
 
 func (c *Cache) WriteoutQueue() *WriteoutQueue {
 	return c.writeoutQueue
+}
+
+// called at every scan-frequency by fileListUpdater in carbonserver. 
+// Iterates over every shard to:
+// - collect the new metric names (append shard.adds map to slice)
+// - replace shard.adds with new empty map
+func (c *Cache) GetRecentNewMetrics() []map[string]struct{} {
+	metricNames := make([]map[string]struct{},shardCount)
+	for i:=0; i<shardCount ; i++ {
+		shard, newAdds := c.data[i], make(map[string]struct{})
+		shard.Lock()
+		currNames := shard.adds
+		shard.adds = newAdds
+		shard.Unlock()
+		metricNames[i] = currNames
+	}
+	return metricNames
 }
--- a/carbonserver/carbonserver.go
+++ b/carbonserver/carbonserver.go
@@ -859,7 +859,10 @@ func (listener *CarbonserverListener) expandGlobs(ctx context.Context, query str
 	var useGlob bool
 
 	// TODO: Find out why we have set 'useGlob' if 'star == -1'
-	if star := strings.IndexByte(query, '*'); strings.IndexByte(query, '[') == -1 && strings.IndexByte(query, '?') == -1 && (star == -1 || star == len(query)-1) {
+	if star := strings.IndexByte(query, '*'); listener.cacheGetRecentMetrics == nil &&
+		strings.IndexByte(query, '[') == -1 &&
+		strings.IndexByte(query, '?') == -1 &&
+		(star == -1 || star == len(query)-1) {
 		useGlob = true
 	}
 	logger = logger.With(zap.Bool("use_glob", useGlob))
@@ -896,23 +899,6 @@ func (listener *CarbonserverListener) expandGlobs(ctx context.Context, query str
 		fallbackToFS = true
 	}
 
-	if useGlob || fallbackToFS {
-		// no index or we were asked to hit the filesystem
-		for _, g := range globs {
-			nfiles, err := filepath.Glob(listener.whisperData + "/" + g)
-			if err == nil && nfiles != nil {
-				files = append(files, nfiles...)
-			} else {
-				// when cache-scan is enabled in the config, fallback to
-				// trigram-searching code when there is no whisper file
-				// to see if the metric exists only in cache
-				if listener.cacheGetRecentMetrics != nil{
-					useGlob = false
-				}
-			}
-		}
-	}
-
 	if fidx != nil && !useGlob {
 		// use the index
 		docs := make(map[trigram.DocID]struct{})
@@ -941,6 +927,19 @@ func (listener *CarbonserverListener) expandGlobs(ctx context.Context, query str
 		}
 
 		sort.Strings(files)
+	}
+
+	// Not an 'else' clause because the trigram-searching code might want
+	// to fall back to the file-system glob
+
+	if useGlob || fallbackToFS {
+		// no index or we were asked to hit the filesystem
+		for _, g := range globs {
+			nfiles, err := filepath.Glob(listener.whisperData + "/" + g)
+			if err == nil {
+				files = append(files, nfiles...)
+			}
+		}
 	}
 
 	leafs := make([]bool, len(files))
--- a/carbonserver/carbonserver.go
+++ b/carbonserver/carbonserver.go
@@ -128,6 +128,13 @@ var statusCodes = map[string][]uint64{
 	"capabilities": make([]uint64, 5),
 }
 
+// interface to retrive retention and aggregation
+// schema from persister.
+type configRetriever interface {
+	MetricRetentionPeriod(string) (int, bool)
+	MetricAggrConf(string) (string, float64, bool)
+}
+
 type responseWriterWithStatus struct {
 	http.ResponseWriter
 	statusCode int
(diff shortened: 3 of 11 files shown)


carbonserver/carbonserver.go
```diff
@@ -591,19 +607,51 @@ func (listener *CarbonserverListener) UpdateMetricsAccessTimesByRequest(metrics
...
 			return
 		case <-tick:
 		case <-force:
 		}
-		listener.updateFileList(dir)
+		if listener.cacheGetRecentMetrics != nil{
+			// cacheMetrics maintains all new metric names added in cache
+			// when cache-scan is enabled in conf
+			newCacheMetricNames := listener.cacheGetRecentMetrics()
+			cacheMetrics = splitAndInsert(cacheMetrics, newCacheMetricNames)
```
bom-d-van, line 648: ~~nits: why not just calling this func in updateFileList, this way you don't have to change func signatures of listener.updateFileList.~~

carbonserver/carbonserver.go
```diff
@@ -826,6 +896,23 @@ func (listener *CarbonserverListener) expandGlobs(ctx context.Context, query str
...
+		for _, g := range globs {
+			nfiles, err := filepath.Glob(listener.whisperData + "/" + g)
+			if err == nil && nfiles != nil {
+				files = append(files, nfiles...)
+			} else {
+				// when cache-scan is enabled in the config, fallback to
+				// trigram-searching code when there is no whisper file
+				// to see if the metric exists only in cache
+				if listener.cacheGetRecentMetrics != nil{
+					useGlob = false
```
bom-d-van: note: trigram index isn't very efficient when it comes to searching full-named metric, i.e. not containing any pattern matching chars (*, [], {}, etc).

And would this causing duplicated search result?

carbonserver/carbonserver.go
bom-d-van: So overtime cacheMetrics would contain all the files and directories of the whole whisper data directory? Probably few hundreds megabytes or more of memory overhead for 4-5 millions of metrics?

cache/cache.go
```diff
@@ -63,6 +63,7 @@ type Shard struct {
 	items            map[string]*points.Points
 	notConfirmed     []*points.Points // linear search for value/slot
 	notConfirmedUsed int              // search value in notConfirmed[:notConfirmedUsed]
+	adds             map[string]struct{} // map to maintain all the new metric names
```
bom-d-van, line 66: not really about this PR, but if we could move to lock free map or just sync.Map for items. I guess the memory overhead from cacheMetrics in fileListUpdater could be saved.

carbonserver/carbonserver.go
```diff
@@ -624,6 +671,24 @@ func (listener *CarbonserverListener) updateFileList(dir string) {
 	}
 
 	metricsKnown := uint64(0)
+
+	// populate index for all the metric names in cache
+	// the iteration takes place only when cache-scan is enabled in conf
+	for fileName := range cacheMetricNames {
```
bom-d-van, line 677: Just a comment: this has a side effect of random order while ranging over maps. So keeping directories in cacheMetricNames is probably unnecessary for trie index.

deploy/go-carbon.conf
```diff
@@ -238,6 +238,12 @@ scan-frequency = "5m0s"
 #  memory usage (by setting both trie-index and trigram-index to true).
 trie-index = false
 
+# This provides the ability to query for new metrics without any wsp files
+# i.e query for metrics present only in cache. Does a cache-scan and
+# populates index with metrics with or without corresponding wsp files.
+# Disabled by default.
+cache-scan = false
```
bom-d-van, line 245: If my comment on extra memory overhead is correct, could you mention the side effect of keeping a copy of all the files/dirs in memory here?

carbonserver/fetchsinglemetric.go
```diff
@@ -96,44 +98,63 @@ func (listener *CarbonserverListener) fetchSingleMetric(metric string, pathExpre
...
+			PathExpression: pathExpression,
+		}
+		cacheData, cerr := listener.fetchFromCache(metric, fromTime, untilTime, &resp)
+		if cerr != nil {
+			atomic.AddUint64(&listener.metrics.RenderErrors, 1)
+			logger.Warn("metric not found even in Cache", zap.Error(err))
+			// Metric has no Whisper file and/or has no datapoints in Cache
+			return response{}, cerr
+		}
+		resp.enrichFromCache(listener, cacheData)
```
bom-d-van, line 147: ~~is this still necessary? Since it's fetched from cache already.~~

never mind, seems you reuse enrichFromCache for actually populating the data.

carbonserver/carbonserver.go
saipraneethyss: That is correct! There is memory overhead involved.

carbonserver/carbonserver.go
saipraneethyss: The intention here is to do a filepath glob first *and then* to fallback to trigram query when `cache-scan` is enabled. This way, for metrics with wsp files on disk, we only hit the filepath glob, and for cache-only metrics, we will also query the trigram index.

deploy/go-carbon.conf
saipraneethyss: Sure. I will update the comment and mention about the memory overhead.

carbonserver/fetchsinglemetric.go
saipraneethyss: Yes. The actual data is populated in the call to enrichFromCache.

carbonserver/carbonserver.go
saipraneethyss: I am not much familiar with the existing trie index implementation and logic. Can you please explain a little more about this?

carbonserver/carbonserver.go
bom-d-van: ~~Files exists on disk could also be indexed in trigram index, right? So if the the globs has miss matches in filepath.Glob, then useGlob is set to false. And there is another `for _, g := range globs ` when querying trigram index bellow?~~

edit: sorry, I was confused. filepath.Glob doesn't return a not found error. https://golang.org/pkg/path/filepath/#Glob

bom-d-van: Actually, if the there is malformed glob pattern in the globs, it could caused duplicated results because now you would range globs twice here and below.

I am not sure if this change is necessary? If you want to force go-carbon to always use trigram index in order to hit the still in-memory new metrics, maybe it's better to just force set useGlob to false?

carbonserver/carbonserver.go
bom-d-van: For trie index, it always create a dir if not exists when inserting a metric. So if dirs didn't come in order anymore, they don't have to be cached.
It doesn't matter much, could just be a small memory savings when both trie and cache-scan are enabled.

carbonserver/carbonserver.go
saipraneethyss: Oh! I see. Agreed. I will change the logic to force set useGlob to false when cache-scan is enabled.

README.md
```diff
@@ -316,6 +316,12 @@ scan-frequency = "5m0s"
 #  memory usage (by setting both trie-index and trigram-index to true).
 trie-index = false
 
+# This provides the ability to query for new metrics without any wsp files
+# i.e query for metrics present only in cache. Does a cache-scan and
+# populates index with metrics with or without corresponding wsp files.
+# Disabled by default.
+cache-scan = false
```
bom-d-van, line 323: > but will lead to increased memory consumption

Need to patch here as well.

bom-d-van:
Other than the last comment about README, lgtm. Thanks for the patch. :)
So should be backward compatible and no behaviour changes if the option is disabled.

README.md
saipraneethyss: Thank you for mentioning this. Missed to add the comment here. Pushed the change!

azhiltsov:
I am going to review and test this in our environment in the nearest week. Sorry for delay.

saipraneethyss:
@azhiltsov, No problem. Thank you for looking into this.

jdblack:
Is there a sense of how long this will take to make it into release?     This issue has become severe enough to us to consider cutting a local package.

azhiltsov:
I am sorry to say, but unfortunately due to a change in priorities I did not have a time to test this in our environment. Currently I do not have estimates when this can be done. 
@lomik what do you think about merging it?

jdblack:
I don't think this PR is the cause of this, but I'm getting panics while trying to run this.

I have seen the same thing in the past when during our occasional tests to try and replace carbonserver with graphite-web.   We  have never gotten to the bottom of it, and always end up going back.

I'll go try a clean build of 0.14 and file a separate  issue for the panic there.


```

[2020-06-05T20:31:25.608Z] ERROR [access] fetch failed {"handler": "render", "url": "/render/?format=pickle&local=1&noCache=1&from=1591389075&until=1591389085&target=SensorProxy.pod01.%2A.meter.proxy.tcp.accept.one-minute&now=1591389085", "peer": "[ipaddressremoved:44134", "format": "pickle", "targets": ["SensorProxy.pod01.*.meter.proxy.tcp.accept.one-minute"], "runtime_seconds": 0.000221996, "reason": "panic during serving the request", "stack": "github.com/lomik/go-carbon/carbonserver.(*CarbonserverListener).renderHandler.func1
\t/Users/jblackwell/go-carbon/_vendor/src/github.com/lomik/go-carbon/carbonserver/render.go:199
runtime.gopanic
\t/usr/local/Cellar/go/1.14.3/libexec/src/runtime/panic.go:969
runtime.panicmem
\t/usr/local/Cellar/go/1.14.3/libexec/src/runtime/panic.go:212
runtime.sigpanic
\t/usr/local/Cellar/go/1.14.3/libexec/src/runtime/signal_unix.go:695
github.com/lomik/go-carbon/carbonserver.(*CarbonserverListener).getExpandedGlobs
\t/Users/jblackwell/go-carbon/_vendor/src/github.com/lomik/go-carbon/carbonserver/find.go:355
github.com/lomik/go-carbon/carbonserver.(*CarbonserverListener).prepareDataProto
\t/Users/jblackwell/go-carbon/_vendor/src/github.com/lomik/go-carbon/carbonserver/render.go:343
github.com/lomik/go-carbon/carbonserver.(*CarbonserverListener).fetchWithCache
\t/Users/jblackwell/go-carbon/_vendor/src/github.com/lomik/go-carbon/carbonserver/render.go:301
github.com/lomik/go-carbon/carbonserver.(*CarbonserverListener).renderHandler
\t/Users/jblackwell/go-carbon/_vendor/src/github.com/lomik/go-carbon/carbonserver/render.go:207
net/http.HandlerFunc.ServeHTTP
\t/usr/local/Cellar/go/1.14.3/libexec/src/net/http/server.go:2012
github.com/lomik/go-carbon/carbonserver.TraceHandler.func1
\t/Users/jblackwell/go-carbon/_vendor/src/github.com/lomik/go-carbon/carbonserver/carbonserver.go:186
github.com/lomik/go-carbon/vendor/github.com/dgryski/httputil.TimeHandler.func1
\t/Users/jblackwell/go-carbon/_vendor/src/github.com/lomik/go-carbon/vendor/github.com/dgryski/httputil/times.go:26
github.com/lomik/go-carbon/vendor/github.com/dgryski/httputil.TrackConnections.func1
\t/Users/jblackwell/go-carbon/_vendor/src/github.com/lomik/go-carbon/vendor/github.com/dgryski/httputil/track.go:40
net/http.HandlerFunc.ServeHTTP
\t/usr/local/Cellar/go/1.14.3/libexec/src/net/http/server.go:2012
net/http.(*ServeMux).ServeHTTP
\t/usr/local/Cellar/go/1.14.3/libexec/src/net/http/server.go:2387
github.com/lomik/go-carbon/vendor/github.com/NYTimes/gziphandler.GzipHandlerWithOpts.func1.1
\t/Users/jblackwell/go-carbon/_vendor/src/github.com/lomik/go-carbon/vendor/github.com/NYTimes/gziphandler/gzip.go:293
net/http.HandlerFunc.ServeHTTP
\t/usr/local/Cellar/go/1.14.3/libexec/src/net/http/server.go:2012
net/http.serverHandler.ServeHTTP
\t/usr/local/Cellar/go/1.14.3/libexec/src/net/http/server.go:2807
net/http.(*conn).serve
\t/usr/local/Cellar/go/1.14.3/libexec/src/net/http/server.go:1895", "error": "runtime error: invalid memory address or nil pointer dereference", "http_code": 500}^@

```

saipraneethyss:
Hi @azhiltsov, @lomik  can you please take a look at this PR and consider merging this?

deniszh:
@saipraneethyss : 
Could you please rebase this PR? 
go-carbon was moved to go-graphite project, so, paths were changed. 
Thanks a lot, sorry for inconvenience!

deniszh:
Last call - @bom-d-van @azhiltsov - any objections to merge this? Thanks for reviewing!

azhiltsov:
No objections. Thanks to everyone involved!

deniszh:
Cool, merging then!

PR #348, opened 2020-07-08
Upgrading go and fpm

deniszh:
Testing Travis-Ci


deniszh:
Ok, only release integration left

PR #364, opened 2020-07-24
Avereha/tracing

avereha:
This change is adding ability to generate traces, send them using jaeger protocol and also log them to stdout if needed for debugging.

For this, I used the opentelemetry-go library, because OpenTracing and OpenCensus have merged to form OpenTelemetry according to https://opentracing.io/.
For the most common requests(find and render) I also added a few tags to the span.

Part of this work, there are two semi-related changes: 
 - I moved the other tracing functions that we had to their own file.
 - It seems that my editor is running 'gofmt' on each save. What do you think, should we make a different MR that just runs 'gofmt' and maybe a check in CI? If this is causing issues I can update my editor configuration to avoid this type of changes.

This change was tested with production traffic on one server.


carbonserver/trace.go
```diff
@@ -0,0 +1,154 @@
...
+					kv.String("host.hostname", fqdn),
+				},
+			}),
+			jaeger.RegisterAsGlobal(),
+			jaeger.WithSDK(&sdktrace.Config{DefaultSampler: sdktrace.AlwaysSample()}),
+		)
+		if err != nil {
+			logger.Fatal("can't initialize jaeger tracing exporter", zap.Error(err))
+		}
+		return flush
```
bom-d-van, line 75: It seems the returned flush func [isn't called](https://github.com/go-graphite/go-carbon/pull/364/files#diff-4c1a6b20ef4151de2ba19edf33e5dabdR471)? Is that intended?

bom-d-van (commented):
Nice work. 
Added a small comment. Also included @deniszh for review.

carbonserver/trace.go
avereha: Nice find, thank you!
I think I fixed this and now we call flush() traces on exit.

deniszh:
I think having `go fmt` and `go mod tidy` in linter is a good idea, but it will require some work, indeed. Let's keep new code with fmt applying for now, but it will be nice to do massive formating change together with linter.

deniszh (approved):

deniszh:
Looking goot to merge, @avereha @bom-d-van ?

bom-d-van:
Yep, lgtm. 👍

Issue #365, opened 2020-07-28
Apply common go linters


deniszh:
It's good to have `go fmt`, `go mod tidy` etc. to be applied / tested on every PR.

dgryski:
Also `go vet`.  I recommend also `ineffassign`, `nilness` and of course `staticcheck`, but getting all these passing will require some code changes.

(While we're listing linters I like: ruleguard and gocritic are also reasonable).

deniszh:
Yes, enabling linters is easy, but for changes I need a bit of time. :P

deniszh:
And (as I lazy) I'm mostly looking for applying tuned [golangci](https://golangci-lint.run/usage/linters/) through [golangci-lint-action](https://github.com/golangci/golangci-lint-action).

deniszh:
Implemented in https://github.com/go-graphite/go-carbon/pull/368

PR #368, opened 2020-09-06
Implementing linters

deniszh:
Fixing #365. Becoming quite tedious task, but it was fun.
Enabling `gofmt`, `goimports` and `gocritic`, together with others [golang-cilint defaults](https://golangci-lint.run/usage/linters/#enabled-by-default-linters) but disabling `errcheck` (it's too noisy IMO).
I avoided fixing unused struct and [S1012](https://staticcheck.io/docs/checks#S1012).


carbonserver/details.go
```diff
@@ -7,19 +7,13 @@ import (
...
-	"unsafe"
 
 	"go.uber.org/zap"
 
 	"github.com/go-graphite/carbonzipper/zipper/httpHeaders"
-	protov2 "github.com/go-graphite/protocol/carbonapi_v2_pb"
 	protov3 "github.com/go-graphite/protocol/carbonapi_v3_pb"
 )
 
-func metricDetailsV2toV3(ptr *map[string]*protov2.MetricDetails) *map[string]*protov3.MetricDetails {
```
deniszh, line 19: /cc @Civil : could you please check - was `metricDetailsV2toV3` needed?

Civil: That's some leftovers from attempts to cheaply convert data from one format to another. So it should be safe to remove.

deniszh: Thanks for checking, @Civil !

avereha (approved):

PR #377, opened 2020-10-20
Upgrade golangci-lint action

deniszh:
to v2.2.1


deniszh:
oh, newer liner is much more strict, need additional PR for fixing all issues

PR #385, opened 2020-10-27
carbonserver: fix incorrect hanlding of couting and indexing files in updateFileList

bom-d-van:
It was introduced since v0.15.3 for the concurrent-index and file-list-cache cache feature.


bom-d-van:
@deniszh thank you!

Issue #395, opened 2021-01-19
[BUG] Values of new metrics not renderable with realtime-index enabled.


timw6n:
**Describe the bug**

Quite a similar situation to that reported in https://github.com/go-graphite/go-carbon/issues/372. We have a large cluster (~3 million metrics) and so make heavy use of the cache, with max-updates-per-second throttled down to keep disk activity within sensible limits. 

Have recently upgraded to 0.15.5, and at that point enabled realtime-index and concurrent-index. At present the issue we're having is that the names of new metrics are visible immediately (and so can be found by Graphite find queries) but the datapoints cannot be retrieved from the cache (render queries) until the whisper files end up on disk. New datapoints on existing metrics pull through from the cache fine.

Logged error seems to be 

```
[...]

PR #396, opened 2021-01-20
trie/realtime: set wspConfigRetriever when realtime index is enabled #395

bom-d-van:


deniszh (approved):

deniszh:
Looks like it's fixing #395. Will merge.

deniszh (approved):

bom-d-van:
Yep, looks so. The deepsource errors aren't actually introduced by this commit and seems to be overlapping with golangci lint but not using the lint ignore/skip comments, so I decided not to address them here.

Thanks @deniszh for following it up.

PR #401, opened 2021-02-06
fix default output path to match package-tree rule

k3a:
package-tree Makefile rule explicitly refers to ./build/$(NAME)-linux-amd64 so make sure default go build uses that path too

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@ SOURCES=$(shell find . -name "*.go")
 all: $(NAME)
 
 $(NAME): $(SOURCES)
-	$(GO) build --ldflags '-X main.BuildVersion=$(BUILD)' $(MODULE)
+	$(GO) build -o ./build/$(NAME)-linux-amd64 --ldflags '-X main.BuildVersion=$(BUILD)' $(MODULE)
 
 debug: $(SOURCES)
 	$(GO) build -ldflags=-compressdwarf=false -gcflags=all='-l -N' $(MODULE)


deniszh:
Hi @k3a ,

Sorry, but that's not how it works. Build for package-tree is done using gox. See https://github.com/go-graphite/go-carbon/blob/master/.travis.yml#L18-L20
I know, it's a bit cryptic and can be simplified.

k3a:
Hi @deniszh,
I saw that gox rule but if you are using gox, why is there the go-carbon (generic) rule?

I am maintaining Arch Linux AUR package of [go-carbon](https://aur.archlinux.org/packages/go-carbon/) (despite not actively using it ATM :P) and I think installing gox during the build process would be quite overkill (and not really a "go style") as it builds for a singe platform only (the one it is compiling on).

Currently I am doing this:

```sh
    make all BUILD="$pkgver"
    mv ./go-carbon ./build/go-carbon-linux-amd64
    make package-tree
```
I can continue doing that, no big deal, it's up to you. ;)

Though previously I was able to do `make all package-tree BUILD="$pkgver"` without that mv sub-step.

deniszh:
> I saw that gox rule but if you are using gox, why is there the go-carbon (generic) rule?

If you want just build a binary instead of a package.

> I can continue doing that, no big deal, it's up to you.

I would like to avoid changing default behaviours. It's used in Dockerfile and probably in many other places. IMO using gox for package build is fine, usually users not doing that and package maintainers can afford extra step.

k3a:
I see, it probably makes sense to keep as is, then. 

Arch Linux does not use deb or rpm.
I can continue doing that middle-step, though. It is just a single binary so it doesn't matter...

Closing the PR, sorry for bothering you.

Issue #400, opened 2021-02-03
[Q] Disabling scan-frequency for Carbonserver


loitho:
Hi there,

excuse the maybe naive question but, we're using go-carbon and are receiving 650 000 Metrics / Node, out of a 4 node cluster (so 2.5 Million metrics) per minute.
One issue we're facing is that the load iops peak tremendously when a scan is triggered to build the index for carbonserver.
There is around 3 750 000 metrics per node.
I then applied the following configuration for Carbonserver : 
```
[carbonserver]
enabled = true
buckets = 10
metrics-as-counters = false
read-timeout = "60s"
write-timeout = "60s"

query-cache-enabled = true
query-cache-size-mb = 0
find-cache-enabled = true

trigram-index = false
scan-frequency = "0m0s"
trie-index = true
```
As you can see, I set *scan-frequency* to 0.
And it's working nicely, my servers don't get choked for 5 minutes (very high load during this period) trying to read all of the files.
(This behavior was happening even when only the trie index was enabled)

And so I was wondering, is this a problem to run the configuration like that ? Considering that I'm still able to have good performance from my cluster compared to before, is there a reason where I should enable this setting back ?

Kind regards,

Thomas

bom-d-van:
Hi @loitho , if scan-frequency is set to 0. No index is built. `trie-index = true` is a no-op. 

It's a trade-off in the current system. Without index, your queries might become slower as it falls back to using filesystem globing, but things should continue to work.

How many memory does your server have? If the memory capacity of the server is big enough, the kernel should be able cache all the file system metadata in memory and you shouldn't have too much io issues caused by scanning directories.

---

Not sure how many people are having with issues with file system scanning. But now with concurrent and realtime indexing support in trie-index, we should be able to support indexing without scanning.

loitho:
Hi @bom-d-van thank you for your quick reply ! I understand, so basically, carbonserver is behaving like a graphite web instance and looking at the whisper file. Which is honnestly still a pretty good thing as it stops me from having to install a graphite-web (nginx + gunicorn) on each of my node.

> if scan-frequency is set to 0. No index is built. trie-index = true is a no-op.

Makes sense, I didn't see any "build index time" on the graph so I assumed so :) 

>How many memory does your server have? if the memory capacity of the server is big enough, the kernel should be able cache all the file system metadata in memory and you shouldn't have too much io issues

Each node has 32 GB of RAM, is there a way to make sure the kernel has put the file system metadata in cache ?
[...]

PR #403, opened 2021-02-25
Updating documentation

loitho:
Hi,
As discussed in #400 , I'm adding a buit of documentation :)
Please tell me If I did something wrong or if you see better improvements to be made

Kind regards,
Thomas

### DIFF
--- a/go-carbon.conf.example
+++ b/go-carbon.conf.example
@@ -3,7 +3,8 @@
 user = "carbon"
 # Prefix for store all internal go-carbon graphs. Supported macroses: {host}
 graph-prefix = "carbon.agents.{host}"
-# Endpoint for store internal carbon metrics. Valid values: "" or "local", "tcp://host:port", "udp://host:port"
+# Endpoint fo
+# r store internal carbon metrics. Valid values: "" or "local", "tcp://host:port", "udp://host:port"
 metric-endpoint = "local"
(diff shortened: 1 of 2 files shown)


go-carbon.conf.example
```diff
@@ -308,10 +333,13 @@ encoding-time = "iso8601"
...
 encoding-duration = "seconds"
 
-[[logging]]
-logger = ""
-file = "stdout"
-level = "error"
-encoding = "mixed"
-encoding-time = "iso8601"
-encoding-duration = "seconds"
+# You can define multiply loggers:
```
deniszh, line 336: Why do you commenting out default logger in example config?

deniszh:
Thanks for this, @loitho , left question.

loitho:
Hi @deniszh , I'm not too familiar with github, where did you leave your questions ? I'm not seeing anything but I might not be looking at the right place

deniszh:
Ah, I forgot to submit 🤦 
sorry

go-carbon.conf.example
loitho: It was commented out in the README.md, so I just copied the modification to the example config. 
There is already a default logger above, is it useful to have a second logger enabled instead of juste saying that it's available if need be ?

deniszh (approved):

PR #405, opened 2021-03-04
Systemd improvements: start after network/fs online and allow more open files

dvanders:
This PR fixes an issue where go-carbon would start before the whisper data dir was mounted, leading to boot issues.
Also network-online.target is the correct thing to be After and Want.
Lastly, allow more open files to enable larger installations OOTB.


dvanders:
Not sure those failures are related.

deniszh:
Indeed, looks like unrelated. I'll merge and fix tests later. Thank for your contribution!

PR #406, opened 2021-03-05
Tests fix

deniszh:
- pining ubuntu
 - removing go 1.13

### DIFF
--- a/carbonserver/carbonserver.go
+++ b/carbonserver/carbonserver.go
@@ -837,7 +837,7 @@ func (listener *CarbonserverListener) updateFileList(dir string, cacheMetricName
 
 	var freeSpace uint64
 	// diskspace can be negative and Bavail is therefore int64
-	if stat.Bavail >= 0 { // nolint:staticcheck skipcq: SCC-SA4003
+	if stat.Bavail >= 0 { // nolint:staticcheck // skipcq: SCC-SA4003
 		freeSpace = uint64(stat.Bavail) * uint64(stat.Bsize)
 	}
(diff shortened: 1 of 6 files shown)


carbonserver/carbonserver.go
```diff
@@ -837,7 +837,7 @@ func (listener *CarbonserverListener) updateFileList(dir string, cacheMetricName
 
 	var freeSpace uint64
 	// diskspace can be negative and Bavail is therefore int64
-	if stat.Bavail >= 0 { // nolint:staticcheck skipcq: SCC-SA4003
+	if stat.Bavail >= 0 { // nolint:staticcheck // skipcq: SCC-SA4003
```
bom-d-van, line 840: Nice catch!

deniszh: That's not me, but golangci-lint 1.38
But still dunno why "TestThrottleChan" failing, it's working on my laptop :D

bom-d-van: The test seems to be a bit unstable. Sometimes it passed,  sometimes it failed.
I think we can try modify the 0.95 to 0.9.

bom-d-van: Or I could re-phase it as the current newThrottleTicker can't meet the [0.95, 1.05] range 100% of the time. :P

deniszh: Strange, for me it's  failing 100% on GHA and passing 100% on my laptop. Trying 90%-110% range now.

deniszh: Ok, working now, merging

Issue #414, opened 2021-06-25
[FR] Returns error or generating error log when render and glob throttling kicks in


bom-d-van:
Based on some feedbacks from @obfuscurity on issue https://github.com/go-graphite/go-carbon/issues/379#issuecomment-868677349

> Adding another note after some recent debugging. Turns out when we upgraded from 0.14.x to 0.15.x, we missed seeing the new max-metrics-rendered setting. The low default (1000) was causing responses to be incomplete, which in turn affected carbonapi transforms like highestAverage(). This made it appear that sets were being sorted e.g. alphanumerically, since go-carbon was only returning a very small subset of the overall request.
> 
> IMHO this setting should have had a very explicit callout in the CHANGELOG. It certainly would have saved us days of debugging.

> 
> I can't really speak to what a sane default might be. I think the issue here is less about the value and more about making sure folks are aware of it during the upgrade cycle. Hopefully it's far enough in the past that it won't matter for most other users.
> 
> But in our case, the one query that surfaced this issue was returning almost 30k metrics. I don't think we have a good way of measuring per-query metric numbers (this would be a useful metric to export imho, either in go-carbon or carbonapi), so for now we're erring on the side of caution and increasing our setting to 100k.

PR #416, opened 2021-07-08
Bug fixes and small refactoring for carbonserver and trie index

bom-d-van:
It's mainly a misc changes and bug fixes. 

Mostly backward-compatible changes except the default values of max-metrics-globbed and max-metrics-rendered.

1. cache-scan: only init Cache.Shard.adds when the feature is enabled

> cache-scan is a feature in go-cabron to enable quicker visibility of new metrics.
> However, if the feature is not enabled, no need to cache the new metrics info
> in Shard.adds map.

2. trie: numerious optimization and bug fixes for concurrent and realtime index

> * Fix incorrect index total_runtime calculation
> * Stop caching new metrics between file list scanning (which mean file list scanning interval should be slower than new metrics flushing to disk)
> * Start new file list scanning on startup right after loading files from cache: to avoid serving old file list cache too long
> * Move down the trie_count_nodes_time codes down to be within the indexing "block"
> * Add a new counter cache.droppedRealtimeIndex (use camelCase to stay consistent with the rest of cache.* metrics)
> * Re-initiating trieIndex when failed to load metrics from cache file
> * Some other minor changes here and there

3. carbonserver: change max-metrics-globbed and max-metrics-rendered defaults (for issue: https://github.com/go-graphite/go-carbon/issues/414)

> The default values might be too low for some of the users and it is a default introduced mainly
> for trie index. This commit change the default to a much bigger value to avoid silently truncated
> query result.

4. carbonserver: change error/debug log destination to access.log for CarbonserverListener.prepareDataProto

### DIFF
--- a/README.md
+++ b/README.md
@@ -362,10 +362,10 @@ max-globs = 100
 fail-on-max-globs = false
 
 # Maximum metrics could be returned by glob/wildcard in find request (currently
-# works only for trie index)
+# works only for trie index) (Default: 10,000,000)
 max-metrics-globbed  = 30000
 # Maximum metrics could be returned in render request (works both all types of
-# indexes)
+# indexes) (Default: 1,000,000)
 max-metrics-rendered = 1000
 
 # graphite-web-10-mode
--- a/go-carbon.conf.example
+++ b/go-carbon.conf.example
@@ -277,10 +277,10 @@ max-globs = 100
 fail-on-max-globs = false
 
 # Maximum metrics could be returned by glob/wildcard in find request (currently
-# works only for trie index)
+# works only for trie index) (Default: 10,000,000)
 max-metrics-globbed  = 30000
 # Maximum metrics could be returned in render request (works both all types of
-# indexes)
+# indexes) (Default: 1,000,000)
 max-metrics-rendered = 1000
 
 # graphite-web-10-mode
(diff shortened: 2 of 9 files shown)


README.md
```diff
@@ -362,10 +362,10 @@ max-globs = 100
 fail-on-max-globs = false
 
 # Maximum metrics could be returned by glob/wildcard in find request (currently
-# works only for trie index)
+# works only for trie index) (Default: 10,000,000)
 max-metrics-globbed  = 30000
```
deniszh, line 366: Maybe we should change values in README too?

go-carbon.conf.example
```diff
@@ -277,10 +277,10 @@ max-globs = 100
 fail-on-max-globs = false
 
 # Maximum metrics could be returned by glob/wildcard in find request (currently
-# works only for trie index)
+# works only for trie index) (Default: 10,000,000)
```
deniszh, line 280: Same for example config

deniszh (approved):

README.md
bom-d-van: Hi @deniszh thanks for the feedback. 

The high default for `max-metrics-globbed` and `max-metrics-rendered` practically means no limit.

I was thinking about keeping the current value in readme and example as suggested values.

The new default are for existing go-carbon users where they usually don't have the config on their production system so that they aren't surprised by the new throttling control.

What do you think of it?

bom-d-van: Hi @deniszh I assume you are ok with keeping the value in readme and conf example files. ;)

But if needed, I can make a separate MR to alter them to the default value.

PR #417, opened 2021-07-19
mod: upgrade to go 1.16 and drops SOURCES from Makefile

bom-d-van:
* The SOURCES list slows down make file due to listing 20k go files when upgrade to 1.16.
  Does not seem a necessary when using go build.
* Specify build -mod=vedor.
* Upgrade go version to 1.16.


bom-d-van:
Hi @deniszh @Civil , I'm merging this changes. It should be a mostly harmless changes and it seems tests are passed.

If there are problems, we can revert or patch it.

deniszh:
Looks good, @bom-d-van , thanks!

PR #419, opened 2021-07-21
ctrie: a bug fix for trigram statNodes

bom-d-van:
It was bug introduced with concurrent trie. The problem are only triggered if using
trie and trigram at the same time.


deniszh:
LGTM

PR #432, opened 2021-09-30
carbonserver: adds /admin/info endpoint for returning internal info

bom-d-van:
currently only cache info are returned. it could be used by other programs like buckytool
to check if the operations are hurting go-carbon and causing disruptions of the
service.

this is mostly an early setup. more stat and info could be added and returned in the future.


bom-d-van:
note: it's based on the quota branch. need to merge quota branch first.

bom-d-van:
Closing this mr and as it's a duplicate of https://github.com/go-graphite/go-carbon/pull/433 (which is based on new origin/master rather than quota.)

PR #434, opened 2021-10-07
carbonserver: exclude directories from indexing

bom-d-van:
We might want to introduce a config for controlling it.


bom-d-van:
Push a different take in this PR: https://github.com/go-graphite/go-carbon/pull/445

PR #435, opened 2021-11-15
carbonserver: introduce new api /metrics/list_query

bom-d-van:
The api call allows us to query metrics under a namespace query. Unlike /metrics/find, it would
traverse down the matched namespaces and returns all the metrics within them.

It is created to enable management tools to check how many metrics are within matched namespaces,
how big are they physically or logically, etc.

Currently only json payload is supported. I am postponing adding protobuf support as I am
not sure how far the api could go.

A few usage examples:

    /metrics/list_query/?format=json&target=sys.app.*&limit=1024 (list 1024 metrics under namespace `sys.app`)
    /metrics/list_query/?format=json&target=sys.app.* (by default, limit is 65536)
    /metrics/list_query/?format=json&target=sys.app.*&stats_only=true (returns stats info only)

### DIFF
--- a/carbonserver/carbonserver.go
+++ b/carbonserver/carbonserver.go
@@ -68,6 +68,8 @@ type metricStruct struct {
 	InfoErrors           uint64
 	ListRequests         uint64
 	ListErrors           uint64
+	ListQueryRequests    uint64
+	ListQueryErrors      uint64
 	DetailsRequests      uint64
 	DetailsErrors        uint64
 	CacheHit             uint64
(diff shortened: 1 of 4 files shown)


deniszh (approved):

deniszh:
@bom-d-van : LGTM, Deepsorce complaint about "Control-coupled functions detected", not sure if fix needed, though.

bom-d-van:
Thanks Denis, I have added `skipcq: RVV-A0005` in the two functions. The lint suggestions don't seem a good alternative for them.

grzkv (approved):

PR #438, opened 2021-11-18
persister: online schema migration

bom-d-van:
Historically, if configs are updated in storage-schemas.conf or storage-aggregation.conf, old
metrics would remain unchanged and requires manual delete or migration using whisper-resize.py, etc.

With this PR, go-carbon gains the capability of performing live config migration.
This features used the APIs introduced in https://github.com/go-graphite/go-whisper/pull/25.

There are three new configs introduced in persister:

```
# Enable online whisper file config migration.
#
# To partially enable default migration for only some matched rules in
# storage-schemas.conf or storage-aggregation.conf, we can set
# online-migration-global-scope = "-" and enable the migration in those config
# files.
#
# online-migration-global-scope can also be set any combination of the 3 rules
# as a csv string like: "xff" or "xff,aggregationMethod"
#
# online-migration = false
# online-migration-rate = 5
# online-migration-global-scope = "xff,aggregationMethod,schema"
```

We can also partially enable migration for matched rules by setting online-migration-global-scope to "-"
and tweak the configs in storage-schemas.conf or storage-aggregation.conf.

persister would also emit metrics for this feature for monitoring:

```
persister.onlineMigration.total (counter, total metrics migrated)
persister.onlineMigration.schema (counter, total metrics that have schema migrated)
persister.onlineMigration.xff (counter, total metrics that have xff migrated)
persister.onlineMigration.aggregationMethod (counter, total metrics that aggregationMethod migrated)
persister.onlineMigration.physicalSizeChanges (counter, total physical disk size changes due to config migration)
persister.onlineMigration.logicalSizeChanges (counter, total logical disk size changes due to config migration)
```

---

edit: re-posted [some docs]((https://github.com/go-graphite/go-carbon/pull/438/commits/71f26e01e32e23aceb5a386ad571a5af0a3e76a9)) here for visibility:

### Online config/schema migration

What is online config/schema migration?

it's a process in go-carbon to make sure that metrics are using the right
configurations, if not, it would migrate it to using the right configs.

In general, metrics are always produced with the right configuration specified in
`storage-aggregation.conf` and `storage-schemas.conf` files. However, occasionally,
we might realized that the aggregation method is incorrect for the newly produced
metrics, or the original retention policy is too long or too short. And if we changed
configs now, it would only apply on the new metrics. In order to make sure that
the old metrics are using the right config, we have a few solutions:

* [whisper-resizepy](https://github.com/graphite-project/whisper#whisper-resizepy), [whisper-set-aggregation-methodpy](https://github.com/graphite-project/whisper#whisper-set-aggregation-methodpy), etc.
* Deleting the old metrics and lose the history data.

Manual approach works fine if the data set is not very large. However, automation is always better.

The `online config/schema migration` feature in go-carbon is created to addresses
this issue.

Although it's a nice feature, there are some caveats and limitations. Go-Carbon
tries keep as much history intact as possible. However, depends on the policy
changes, it might not always be possible to do so. Please be careful with the
migration.

**Notable caveats**:

* When applying schema/retention changes, history is only copied if the resolution is the same (See some examples bellow).
* When applying aggregation method changes, histories are not changed.
* It's important to access the file size changes before applying new configs.

Expected changes:

| Policy Changes  | Temporary File  | History | File Size Changes  |
|---|---|---|---|
| XFiles Factor  | Not Needed  | No Changes  |  No Changes |
| Aggregation Method  | Not Needed | No Changes | No Changes |
| Schema Changes  | Required  | Depends  | Depends  |

Some illustrations of schema changes:

| Original Config | New Config | History | File Size Changes  |
|---|---|---|---|
|`1s:2d,1m:31d,1h:2y` | `1s:2d,1m:62d,1h:2y`|Copied in All Archives|Increase|
|`1s:2d,1m:62d,1h:2y` | `1s:2d,1m:31d,1h:2y`|Copied in All Archives, and history in the second archive (`1m:62d`) is truncated to fit the new retention |Reduce|
|`1s:2d,1m:31d` | `1s:2d,5m:31d`|History in first archive is copied, history in the second archive is dropped because the resolution is not the same|Reduce|
|`1s:2d,5m:31d` | `1s:2d,1m:31d` |History in first archive is copied, history in the second archive is dropped because the resolution is not the same|Increase|

---

edit on May 4th:

Updated go-whisper to include bug fixes in cwhisper: https://github.com/go-graphite/go-whisper/pull/32

Also introduce tool/persister_configs_differ for convenience (although it might be better if we turn it into a go-carbon subcommand for greater convenience).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@
 * Introducing `persiter.oooDiscardedPoints` metric #463
 * Fixing OOO discard metric overflow #464
 * refactoring `persiter.oooDiscardedPoints` metric #TBD
+* __Online Schema/Aggregation Migration__ #438
+* tool/persister_configs_differ and -check-policies flag in go-carbon #438
+
 ##### version 0.16.2
 * Another attempt to fix issues with release upload #449
 
--- a/README.md
+++ b/README.md
@@ -155,8 +155,8 @@ remove-empty-file = false
 #
 # To partially enable default migration for only some matched rules in
 # storage-schemas.conf or storage-aggregation.conf, we can set
-# online-migration-global-scope = "-" and enable the migration in those config
-# files.
+# online-migration-global-scope = "-" and enable the migration in the config
+# files (more examples in deploy/storage-aggregation.conf and deploy/storage-schemas.conf).
 #
 # online-migration-global-scope can also be set any combination of the 3 rules
 # (xff,aggregationMethod,schema) as a csv string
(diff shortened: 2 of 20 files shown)


bom-d-van:
Adds a bit [more docs](https://github.com/go-graphite/go-carbon/pull/438/commits/71f26e01e32e23aceb5a386ad571a5af0a3e76a9) about this feature.

grzkv (approved):

Issue #436, opened 2021-11-16
[BUG] panic in render `"error":"interface conversion: interface {} is nil, not carbonserver.fetchResponse"}`


gburek-fastly:
**Describe the bug**
Panic in render function when query cache returns a nil

**Logs**
```
{"level":"ERROR","timestamp":"2021-11-12T08:35:03.874Z","logger":"access","message":"panic recovered","handler":"render","url":"/render/?format=carbonapi_v3_pb","peer":"127.0.0.1:52768","targets":["fake.reports.detail.fake.*.*.{bereq_header_bytes,bereq_body_bytes}"],"format":"carbonapi_v3_pb","carbonapi_uuid":"af5b3450-26e5-4af5-a61f-097c42f60a1b","carbonzipper_uuid":"af5b3450-26e5-4af5-a61f-097c42f60a1b","stack":"github.com/go-graphite/go-carbon/carbonserver.(*CarbonserverListener).renderHandler.func1
	/home/travis/gopath/src/github.com/go-graphite/go-carbon/carbonserver/render.go:195
runtime.gopanic
	/home/travis/.gimme/versions/go1.15.5.linux.amd64/src/runtime/panic.go:969
runtime.panicdottypeE
	/home/travis/.gimme/versions/go1.15.5.linux.amd64/src/runtime/iface.go:261
github.com/go-graphite/go-carbon/carbonserver.(*CarbonserverListener).fetchWithCache
	/home/travis/gopath/src/github.com/go-graphite/go-carbon/carbonserver/render.go:321
github.com/go-graphite/go-carbon/carbonserver.(*CarbonserverListener).renderHandler
	/home/travis/gopath/src/github.com/go-graphite/go-carbon/carbonserver/render.go:209
net/http.HandlerFunc.ServeHTTP
	/home/travis/.gimme/versions/go1.15.5.linux.amd64/src/net/http/server.go:2042
github.com/go-graphite/go-carbon/carbonserver.TraceHandler.func1
	/home/travis/gopath/src/github.com/go-graphite/go-carbon/carbonserver/trace.go:140
github.com/dgryski/httputil.TimeHandler.func1
	/home/travis/gopath/src/github.com/go-graphite/go-carbon/vendor/github.com/dgryski/httputil/times.go:26
github.com/dgryski/httputil.TrackConnections.func1
	/home/travis/gopath/src/github.com/go-graphite/go-carbon/vendor/github.com/dgryski/httputil/track.go:40
[...]

PR #442, opened 2021-12-10
Fix carbonserver render error race

emadolsky:
It is a simple fix to https://github.com/go-graphite/go-carbon/issues/436.

Whenever there are multiple render requests approaching for the
same records, and the proceeding request gets an error, waiting
requests should get an error too.

It is a bit different from #431 because `listener.prepareDataProto` handles the 404 case and returns an empty response which will be handled correctly when it proceeds. So handling the error case would probably be the solution.

### DIFF
--- a/carbonserver/render.go
+++ b/carbonserver/render.go
@@ -314,12 +314,16 @@ func (listener *CarbonserverListener) fetchWithCache(ctx context.Context, logger
 			} else {
 				item.StoreAndUnlock(response)
 			}
-		} else {
+		} else if res != nil {
 			logger.Debug("query cache hit")
 			atomic.AddUint64(&listener.metrics.QueryCacheHit, 1)
 
 			response = res.(fetchResponse)
 			fromCache = true
+		} else {
+			// Whenever there are multiple requests approaching for the same records,
+			// and the proceeding request gets an error, waiting requests should get an error too.
+			err = fmt.Errorf("invalid cache record for the request")
 		}
 	} else {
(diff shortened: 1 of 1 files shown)


deniszh (approved):

deniszh:
Could you please check the fix, @bom-d-van ?

bom-d-van:
yep, overall it looks good to me.

There is a DeepSource failure, @emadolsky could you check if you can fix it?
(not sure what's going on in golangci-lint.)

deniszh:
Not sure about DeepSource - it complains about unused metric parameter - https://deepsource.io/gh/go-graphite/go-carbon/run/617e9c52-2797-47f7-90f9-e9b47844eb47/go/RVV-B0012 but lint error is clear, though - https://go-critic.com/overview.html#ifElseChain-ref 
IMO replace if chain with switch case is good suggestion overall.

emadolsky:
The metric name variable was not used in the function. The only use could be logging it. Also we could just get rid of it. But I was not sure which to choose. What do you think?

bom-d-van:
> The metric name variable was not used in the function. The only use could be logging it. Also we could just get rid of it. But I was not sure which to choose. What do you think?

logging seems good for me. approving it and re-running the jobs.

bom-d-van:
[The CodeQL ones](https://github.com/go-graphite/go-carbon/pull/442/checks?check_run_id=4505568154) look like the reaction of golang linter to the log4j incident. I'm reviewing them, probably gonna mark them as `won't fix` or `false positive`. please let me know if you disagree.

Issue #202, opened 2017-09-14
go-carbon can't save metrics on disk under high go-trigram.intersect load


azhiltsov:
With loose settings of
max-globs = 1000 and high rate of requests (find+render)
the go-trigram.intersect consumes all CPU cores and prevents metrics from flushing to disk
![profile001](https://user-images.githubusercontent.com/8709761/30425369-41d8ee26-9949-11e7-88d5-19e979557562.png)
.
perf top -p PID output
```
Samples: 221K of event 'cycles:ppp', Event count (approx.): 138588923899                                             
Overhead  Shared Object     Symbol                                                                                   
  95.77%  go-carbon         [.] github.com/lomik/go-carbon/vendor/github.com/dgryski/go-trigram.intersect
   0.35%  go-carbon         [.] runtime.semrelease
   0.22%  go-carbon         [.] runtime.memclrNoHeapPointers
   0.15%  go-carbon         [.] runtime.mallocgc
   0.11%  [kernel]          [k] change_protection_range
   0.10%  [kernel]          [k] find_vma
   0.09%  [unknown]         [k] 0x000000000078b8d5
   0.08%  [kernel]          [.] irq_return
   0.08%  go-carbon         [.] runtime.mapaccess2_faststr
   0.06%  go-carbon         [.] runtime.memeqbody
   0.06%  go-carbon         [.] runtime.mapaccess2_fast32
   0.06%  go-carbon         [.] github.com/lomik/go-carbon/cache.(*Cache).Add
   0.06%  [kernel]          [k] do_numa_page
   0.05%  go-carbon         [.] runtime.memmove
   0.04%  [kernel]          [k] _raw_spin_lock
   0.04%  [kernel]          [k] handle_mm_fault
   0.03%  go-carbon         [.] runtime.heapBitsSetType
```
@Civil

dgryski:
:(

Civil:
We've triggered this problem again but now with maxGlobs=100 but 20M metrics per server. Symptoms and trace pretty much the same.

Maybe we should add some special cases, store metrics also in radix tree and in case match is too broad, fallback to fullscan over radix?

dgryski:
[...]

PR #443, opened 2021-12-14
carbonserver: introduce new max-inflight-requests and no-service-when-index-is-not-ready configs

bom-d-van:
carbonserver.max-inflight-requests:

    Maximum inflight requests allowed in go-carbon. Default is 0 which means no limit.
    If limit is reached, go-carbon/carbonserver returns 429 (Too Many Requests).
    This option would be handy as if there are too many long running requests piling up,
    it would slows down data ingestion in persister and slows down index building
    (especially during restart).

carbonserver.no-service-when-index-is-not-ready:
  
    Returns 503 (Service Unavailable) when go-carbon is still building the first
    index cache (trie/trigram) after restart. With trie or trigram index,
    carbonserver falls back to filepath.Glob when index is not ready. And
    filepath.Glob is expensive and not scaled enough to support some heavy
    queries (manifested as high usage and frequent GC). Thus it is possible that
    the read requests is so heavy that it not only slows down index building, it
    also hinders persister from flushing down data to disk and causing cache.size
    overflow.
  
    By default, it is disabled.
  
    It is recommend to enable this flag together with "file-list-cache" (depends on
    the number of metrics, indexing building with "file-list-cache" is usually
    much faster than re-scanning the whole file tree after restart).

These two configs could be used to stop carbonserver from starving persister
and indexing goroutines.

### DIFF
--- a/README.md
+++ b/README.md
@@ -354,6 +354,29 @@ concurrent-index = false
 # (EXPERIMENTAL)
 realtime-index = 0
 
+# Maximum inflight requests allowed in go-carbon. Default is 0 which means no limit.
+# If limit is reached, go-carbon/carbonserver returns 429 (Too Many Requests).
+# This option would be handy as if there are too many long running requests piling up,
+# it would slows down data ingestion in persister and slows down index building
+# (especially during restart).
+max-inflight-requests = 0
+
+# Returns 503 (Service Unavailable) when go-carbon is still building the first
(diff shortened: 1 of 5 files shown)


bom-d-van:
Relevant tickets: 
* https://github.com/grobian/carbonserver/issues/8
* https://github.com/go-graphite/go-carbon/issues/202

PR #444, opened 2021-12-26
Pipeline improvements v2

deniszh:
- Upgrade golangci-lint, fix new lint issues


deniszh:
Still not clear why packages doing this. :(
Will revert.

PR #446, opened 2022-02-07
carbonserver/quota: throughput racy counter fixes and refactoring

bom-d-van:
This PR includes two fixes. Most of the observable fix is in the first commit. The second commit is more subtle and more about refactoring.

* carbonserver: fix broken/untimely quota usage report/reset logics

  The usageRefreshTimeout implementation in updateFileList is quite broken as it produces
  report metric gaps from time to time.
  
  And the listener.refreshQuotaAndUsage call after updateFileList in fileListUpdater loop
  also causes incorrect quota usage report.

* carbonserver/quota: fix racy throughput usage collection and throttling, with some refactorings

  The original throughput usage collection and reset process are done though creating a new
  throughput counter, which might be racy as report and reset are done in parallel. This commit
  fixes it by resetting the same usage counter rather than creating new variables every time
  during usage report.
  
  At the same time, throughput reset are done in intervals (specified by carbonserver.quota-usage-report-frequency).
  This means that there might be latency between resets and causing unnecessary throttling
  when quota is saturated. This commit fixes it by checking against quota during the whole
  range rather than just use fixed quota (similar to sliding window).
  
  Include some tests for checking the changes above.
  
  The rest of the changes are mainly refactoring, like renames and more comments.

### DIFF
--- a/carbonserver/carbonserver.go
+++ b/carbonserver/carbonserver.go
@@ -698,6 +698,10 @@ uloop:
 
 			continue uloop
 		case m := <-listener.newMetricsChan:
+			// listener.newMetricsChan might have high traffic, but
+			// in theory, there should be no starvation on other channels:
+			// https://groups.google.com/g/golang-nuts/c/4BR2Sdb6Zzk (2015)
+
 			fidx := listener.CurrentFileIndex()
 			if listener.trieIndex && listener.concurrentIndex && fidx != nil && fidx.trieIdx != nil {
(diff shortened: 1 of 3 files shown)


bom-d-van:
The end result: the root throughput usage metrics is matching tcp.metricsReceived more accurately and no more jumps.

Screenshot 2022-02-07 at 10 48 22

deniszh (approved):

PR #456, opened 2022-03-18
Makefile: revert .SHELLFLAGS changes

bom-d-van:
We are having errors running `make clean && make` due to the flags on linux, not sure the actual cause,
but by removing it, it works both on linux and osx:

```
$ make clean && make
bash: - : invalid option
bash: - : invalid option
rm -f go-carbon build/* *deb *rpm
bash: - : invalid option
make: *** [clean] Error 1
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,6 @@ MODULE:=github.com/go-graphite/go-carbon
 
 SHELL := bash
 .ONESHELL:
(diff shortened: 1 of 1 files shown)


deniszh:
Probably related to bash or make versions. Nvm, it's only failsafe precautions. Maybe we should also remove `SHELL=bash`, let's check later

PR #462, opened 2022-04-19
vendor: update golang.org/x/sys to resolve 1.18 build issues on osx

bom-d-van:
Build is failing on osx due to the following error:

```
vendor/golang.org/x/sys/unix/syscall_darwin.1_13.go:29:3: //go:linkname must refer to declared function or variable
vendor/golang.org/x/sys/unix/zsyscall_darwin_arm64.1_13.go:27:3: //go:linkname must refer to declared function or variable
vendor/golang.org/x/sys/unix/zsyscall_darwin_arm64.1_13.go:40:3: //go:linkname must refer to declared function or variable
vendor/golang.org/x/sys/unix/zsyscall_darwin_arm64.go:28:3: //go:linkname must refer to declared function or variable
vendor/golang.org/x/sys/unix/zsyscall_darwin_arm64.go:43:3: //go:linkname must refer to declared function or variable
vendor/golang.org/x/sys/unix/zsyscall_darwin_arm64.go:59:3: //go:linkname must refer to declared function or variable
vendor/golang.org/x/sys/unix/zsyscall_darwin_arm64.go:75:3: //go:linkname must refer to declared function or variable
vendor/golang.org/x/sys/unix/zsyscall_darwin_arm64.go:90:3: //go:linkname must refer to declared function or variable
vendor/golang.org/x/sys/unix/zsyscall_darwin_arm64.go:105:3: //go:linkname must refer to declared function or variable
vendor/golang.org/x/sys/unix/zsyscall_darwin_arm64.go:121:3: //go:linkname must refer to declared function or variable
vendor/golang.org/x/sys/unix/zsyscall_darwin_arm64.go:121:3: too many errors
```

Upgrading sys package resolves the issue.
Discussions: https://github.com/golang/go/issues/51706

PR #463, opened 2022-04-26
Introducing `persiter.oooDiscardedPoints` metric

deniszh:
- also contains update for golang.org/x/sys/unix in vendor dir for fix build using Go 18 in Darwin (superceding https://github.com/go-graphite/go-carbon/pull/462)
- I also commented out debug logging for rebuilding writeout queue, it's running every 100 ms and polluting all debug output.


bom-d-van (approved):
lgtm, thanks!

PR #469, opened 2022-06-02
carbonserver: introducing request-timeout, heavy-glob-query-rate-limiters and api-per-path-rate-limiters for read traffic regulation

bom-d-van:
Three new types of read path/traffic control configs are introduced in this commit:

* request-timeout: it is designed to control how long each api call in
  carbonserver can run. The existing timeouts like read, write and idle are for
  http.Server, but this new timeout is for each API call in carbonserver.

* heavy-glob-query-rate-limiters are relatively narrow controls against queries
  that might causes high cpu and memory consumption due to matching over too
  many metrics or nodes at the same time, queries
  like: `*.*.*.*.*.*.*.*.*keyword*`. For these types of queries, trigram might
  be able to handle it better, but for trie and filesystem glob, it's might be too
  expensive when the index tree is large.

* api-per-path-rate-limiters are used for strict api call rate limiting. All
  registered API paths (see carbonserver.Listen for a full list) can be
  controlled separately with it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -338,6 +338,8 @@ metrics-as-counters = false
 # Read and Write timeouts for HTTP server
 read-timeout = "60s"
 write-timeout = "60s"
+# Request timeout for each API call
+request-timeout = "60s"
 # Enable /render cache, it will cache the result for 1 minute
 query-cache-enabled = true
 # 0 for unlimited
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -449,6 +451,64 @@ internal-stats-dir = ""
...
+#     "/metrics/details/"
+#     "/render/"
+#     "/info/"
+#     "/forcescan"
+#     "/admin/quota"
+#     "/admin/info"
+#     "/robots.txt"
+#     ...
+#
+# When max-inflight-requests is set to 0, it means instant rejection.
```
timtofan, line 496: dup?

bom-d-van: kinda but intentional. they are for different rate limiters so decided to duplicate it for readability.

PR #470, opened 2022-06-08
carbonserver: introduce file list cache v2

bom-d-van:
This commit introduces flc v2, the main purpose is to have better quota support during restart.
Mainly for saving physical and logical size info in the flc.

Users can transparently switch from v1 to v2 and vice versa.

More context on quota friendly: with flc v1, when go-carbon is restarted, it would create a trie index
tree using the logical size as physical size for metrics and this would potentially temporary high
physical size accounting and leads to unnecessary throttling if physical size quota is enabled.

With flc v2, go-carbon would keep physical, logical, data points, and first_seen_at timestamp in the 
cache file and use that for rebuilding the trie index after restart. This would avoids temporary incorrect
physical size accounting issues with flcv1.

### DIFF
--- a/README.md
+++ b/README.md
@@ -377,7 +377,26 @@ trie-index = false
 # most of the indexing time if it contains a high number of metrics (10
 # - 40 millions). go-carbon only reads the cached file list once after
 # reboot and the cache result is updated after every scan. (EXPERIMENTAL)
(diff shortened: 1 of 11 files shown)


bom-d-van:
image

PR #472, opened 2022-06-20
protocol: upgrade to the latest version

bom-d-van:
No functionality changes, mostly just updating protocol package and related packages. Mass renaming functions for vt api.


Civil:
You might want to test performance after the change. While VT have a promising performance results (and it might be worth to regen proto library with a newer version of VT), but library still behaves differently.

bom-d-van:
> You might want to test performance after the change. While VT have a promising performance results

nice, will check it out while testing and deploying.

> but library still behaves differently.

oh, it it just the api part, right? 

just in case, it's backward compatible in terms of marshal/unmarshal result, right?

Civil:
Sorry, I meant that approach was different and that might have unexpected effect on memory or cpu consumption.

I would also see if updating protocol repo to use newest version of it gives any benefits

Civil:
About marshal/unmarshal - that is guaranteed by the fact that it's protobuf.

I also used one of the first public versions to generate the code initially, and I think since then they've implemented UnmarshalVT that doesn't use reflection and should be much more efficient than just plain Unmarshal (which uses reflection). That's why I've suggested to update protocol repo to use newest release of VT and regenerate everything with that before actually using it.

Other than that - it supports APIv2 of protobuf and package itself is supported (gogoproto is kinda abandonware now).

emadolsky:
@Civil - We've had a test on our systems and no unexpected results were observed. So I suppose we can merge this.
FYI: @bom-d-van

PR #476, opened 2022-07-04
CarbonV2 gRPC streaming render

emadolsky:
This PR

- adds gRPC listener to carbonserver
- brings CarbonV2 gRPC services into play
- implements streaming gRPC Render functionality

This is manually tested in container environment.

Notes:

- It is derived from @bom-d-van's [protocol/upgrade](https://github.com/go-graphite/go-carbon/tree/protocol/upgrade). So, it should not be merged before #472.
- Only Render is implemented. So we either need to have a hybrid usage in front-end, or we need to implement other RPCs too (e.g. Find).


carbonserver/render.go
```diff
@@ -332,6 +354,91 @@ func (listener *CarbonserverListener) fetchWithCache(ctx context.Context, logger
...
+	metricNames := make([]string, len(metricMap))
+	i := 0
+	for k := range metricMap {
+		metricNames[i] = k
+		i++
+	}
+	return metricNames
+}
+
+func (listener *CarbonserverListener) prepareDataProtoStream(ctx context.Context, targets map[timeRange][]target, expandedGlobs []globs, responseChan chan *protov2.FetchResponse) {
```
bom-d-van: this function seems to duplicate some logics from prepareDataProto, any chance of merging them? can stream be a separate format for prepareDataProto or something?

bom-d-van: hmm, sorry, I think I mis-read it. please ignore.

bom-d-van: you know what, I changed my mind again (orz), it seems possible and reasonable to just support stream in `prepareDataProto`?

emadolsky: Well how about having this PrepareDataProtoStream function for protov2 and have the PrepareDataProto function wrapping PrepareDataProtoStream and eliminate the channel and return the final results on protov2 case?
I should check the code again to see the possibility and effort but just saying for now.

bom-d-van: yep, mostly out of the intention to have less duplicated logics for future changes as we might miss stream if we fix problems in `PrepareDataProto` or vice versa.

carbonserver/render.go
```diff
@@ -565,3 +674,130 @@ func (listener *CarbonserverListener) fetchDataPB(metric string, files []string,
...
+				zap.Stack("stack"),
+				zap.Any("error", r),
+				zap.Int("http_code", http.StatusInternalServerError),
+			)
+			rpcErr = status.New(codes.Internal, "Panic occured, see logs for more information").Err()
+		}
+	}()
+
+	// TODO: implementing cache?
+	fromCache := false
```
bom-d-van, line 659: not a request for change but just an extra note: would need to pay attention to request latency and io when we started testing streaming on prod as it's one of those uncertain bits we have atm.

emadolsky: Absolutely

carbonserver/render.go
emadolsky: 👍

emadolsky: I made some changes. Could you please check it out?

carbonserver/render.go
```diff
@@ -355,20 +367,26 @@ func (listener *CarbonserverListener) prepareDataProto(ctx context.Context, logg
...
+	return metricGlobMap
+}
 
-	var metrics []string
+func (listener *CarbonserverListener) prepareDataStream(ctx context.Context, format responseFormat, targets map[timeRange][]target, metricGlobMap map[string]globs, responseChan chan response) {
+	defer close(responseChan)
 	for tr, ts := range targets {
 		for _, metric := range ts {
+			select {
+			case <-ctx.Done():
```
bom-d-van, line 386: should we check the context errors like [this](https://github.com/go-graphite/go-carbon/blob/quota/realtime-usage/carbonserver/carbonserver.go#L1248-L1253) here?

carbonserver/render.go
bom-d-van: thanks. can't really spot new problems now other than nitpickings, should be good.

bom-d-van:
there are a few deep source check errors, can you check try resolving it or silence it with `// skipcq: ` comment? tx.

carbonserver/render.go
emadolsky: How about having the check outside of prepareDataStream?
Like after reading from the channel, we can check the context errors and act accordingly.

bom-d-van: yep, no objection if it could give us insight of why the requests return early. 

but it seems there were no ctx.Err check in the non-stream logics to begin with? not a hard requirement for this pr.

emadolsky: Yeah the older implementation didn't respect context deadline apparently. However, I added the check. Could you please take a look?

bom-d-van: yep. lgtm. tx.

bom-d-van:
after you mark this pr as ready for review, I guess I can approve it.

just a simple recap and being a paranoid in case I missed it, no backward-incompatibilities are expected, right?

go-carbon.conf.example
```diff
@@ -426,6 +426,12 @@ stats-percentiles = [99, 98, 95, 75, 50]
 # path = "/metrics/list_query/"
 # max-inflight-requests = 3
 
+# carbonserver.grpc is the configuration for listening for grpc clients.
+# [carbonserver.grpc]
+# enabled = true
+# listen = ":7004"
```
bom-d-van, line 433: oh, can you also add that only `/render` is currently supported, and copy this change to [the readme configuration section](https://github.com/go-graphite/go-carbon#configuration) as well? tx.

emadolsky: OK. But why are we maintaining another configuration example inside README? We can just put a link to the example configuration instead, can't we?

emadolsky:
> after you mark this pr as ready for review, I guess I can approve it.

Excellent! I will mark it as ready for review as soon as https://github.com/go-graphite/protocol/pull/12 is merged and the lib ref here is updated. Can you check that out as well?

> just a simple recap and being a paranoid in case I missed it, no backward-incompatibilities are expected, right?

That's right. The old proto messages stay the same (new messages and service are added), old API works in the same way as before alongside with the new gRPC implementation.

go-carbon.conf.example
bom-d-van: maybe. for this pr, let's stick to the duplication. a little bit of copy doesn't hurt. ;)

I'm doing it mainly because they were there, so I just keep them as up-to-date and matching as possible.

In readme it improves visibility of new features, in example conf file it makes it easy for people to just use it. but it's a random opinion.

bom-d-van:
> Excellent! I will mark it as ready for review as soon as https://github.com/go-graphite/protocol/pull/12 is merged and the lib ref here is updated. Can you check that out as well?

I have approved the protocol pr. let's give Vladimir a few hours to see if he has opinions there.

bom-d-van (approved):

PR #481, opened 2022-07-20
Docker build from local copy, expands /var/lib/graphite/ paths in Doc…

flucrezia:
Using hardcoded version of go-carbon within the Dockerfile prevents a correct workflow for Github Actions or any other tool using Docker images because the code is always compiled using the static version and thus conflicting with image tags.

Plus, it expands _/var/lib/graphite/_ subdirectories within the Dockerfile since the default shell is sh and not bash.


deniszh:
Hi @flucrezia 

Make sense. Thanks! 
Merging.

PR #482, opened 2022-07-20
Add stats for find and render requests

auguzun:
Now each **/find** produces read_hits metrics and each **/render** request produce **read_hits, read_bytes**.  During refreshUsage execution these metrics are sent and flushed.

### DIFF
--- a/carbonserver/render.go
+++ b/carbonserver/render.go
@@ -5,7 +5,9 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+
 	"github.com/go-graphite/go-whisper"
+
 	"io/ioutil"
 	"math"
 	"net/http"
--- a/carbonserver/render.go
+++ b/carbonserver/render.go
@@ -5,9 +5,6 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-
-	"github.com/go-graphite/go-whisper"
-
 	"io/ioutil"
 	"math"
 	"net/http"
@@ -23,6 +20,7 @@ import (
 	"google.golang.org/grpc/status"
 
 	"github.com/go-graphite/carbonzipper/zipper/httpHeaders"
+	"github.com/go-graphite/go-whisper"
 	grpcv2 "github.com/go-graphite/protocol/carbonapi_v2_grpc"
 	protov2 "github.com/go-graphite/protocol/carbonapi_v2_pb"
 	protov3 "github.com/go-graphite/protocol/carbonapi_v3_pb"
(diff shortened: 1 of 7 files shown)


carbonserver/trie.go
```diff
@@ -467,6 +467,13 @@ func (tn *trieNode) setChild(i int, n *trieNode) {
 	)
 }
 
+func (tn *trieNode) incrementFindMetric() {
+	if tn.file() {
+		meta := tn.meta.(*fileMeta)
+		atomic.StoreInt64(&meta.readHits, meta.readHits+1)
```
bom-d-van: we should use [atomic.AddInt64](https://pkg.go.dev/sync/atomic#AddInt64) here?

carbonserver/trie.go
```diff
@@ -1936,7 +1943,7 @@ func (ti *trieIndex) refreshUsage(throughputs *throughputQuotaManager) (files ui
 			cur.logicalSize += cur.node.meta.(*fileMeta).logicalSize
 			cur.physicalSize += cur.node.meta.(*fileMeta).physicalSize
 			cur.dataPoints += cur.node.meta.(*fileMeta).dataPoints
-
+			cur.readHits += cur.node.meta.(*fileMeta).readHits
```
bom-d-van: we also need to use atomic.LoadInt64 here too because unlike other metas, these is concurrent write.

carbonserver/trie.go
```diff
@@ -467,6 +467,13 @@ func (tn *trieNode) setChild(i int, n *trieNode) {
 	)
 }
 
+func (tn *trieNode) incrementFindMetric() {
+	if tn.file() {
+		meta := tn.meta.(*fileMeta)
```
bom-d-van, line 473: just to be on safe side, let's also check if meta isn't nil:

```
meta, ok := tn.meta.(*fileMeta)
if !ok { return }
```

carbonserver/trie.go
bom-d-van: one more thing, maybe let's also go for making readHits a gauge rather than ever-increasing counter, like we are doing for throttled and throughput?

```
readHits := atomic.LoadInt64(&cur.node.meta.(*fileMeta).readHits)
atomic.AddInt64(&cur.node.meta.(*fileMeta).readHits, -readHits)
```

carbonserver/trie.go
bom-d-van: sorry, scratch this, I think it's not really needed as we always create file trieNode with fileMeta.

carbonserver/render.go
```diff
@@ -5,6 +5,9 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+
+	"github.com/go-graphite/go-whisper"
```
bom-d-van: nit-picking: probably better move it down bellow the std libraries.

carbonserver/trie.go
```diff
@@ -467,6 +468,19 @@ func (tn *trieNode) setChild(i int, n *trieNode) {
...
+		atomic.AddInt64(&meta.readHits, 1)
+	}
+}
+func (tn *trieNode) incrementFindBytesMetric(bytesNumber int64) {
```
bom-d-van: nit-picking: `incrementRenderBytesMetric` might be better as only `/render/` api call return metric data.

bom-d-van (approved):
left 2 more nit-pickings, not a blocker. approved! 👍

bom-d-van (approved):

PR #492, opened 2022-08-31
Add simple cache for grpc render

emadolsky:


deniszh (approved):
Discussed offline, because I didn't really understand how streaming makes sense together with caching, so - it doesn't, but we need to implement caching now to fix performance issue and fix it later.

PR #499, opened 2022-09-30
Solved problem with different number of metrics in trie index and on disk

auguzun:
Problem: during go carbon server provisioning if creation throttling was on we had different number of metrics in trie index and on disk. It happened because we added throttled file to the index, instead of deleting it.
Solution: 
I moved creation throttle logic to [trie.go](https://github.com/go-graphite/go-carbon/pull/499/files#diff-e63d6705c67a01f129c300baba9e3c7b4ac1900ec2062497fcc32b4339661ef0R2317) as well as related metrics. 
So write now to metric goes first to [Add](https://github.com/go-graphite/go-carbon/blob/aguzun/throttle_rework/cache/cache.go#L263), here it may be throttled or may be passed thru to [store function](https://github.com/go-graphite/go-carbon/pull/499/files#diff-c1f7a50d4c8070bc973b5aa894a2ab463b1a512629a0b6b99ad3077e19f49191L245).


emadolsky (approved):

carbonserver/trie.go
```diff
@@ -2221,17 +2230,24 @@ func (ti *trieIndex) getNodeFullPath(node *trieNode) string { // skipcq: SCC-U10
...
+	case keep, open := <-ti.maxCreatesTicker.C:
+		if keep || !open {
+			return false
+		}
+		break
+	default:
+		break
 	}
+	atomic.AddUint64(&ti.throttledCreates, 1)
+	ti.logger.Error("metric creation throttled",
```
azhiltsov, line 2244: the amount of generated logs will correlate with the incoming traffic. May I suggest replacing this by a Prometheus counter and removing the logs?

PR #521, opened 2023-02-07
fix(carbonserver): find http/grpc - fix metrics_found metric

timtofan:


emadolsky (approved):
Approved. But just as an opinion, I think we might be able to call `listener.populateMetricsFoundStat` even when there is a loop which does something else as well. It's cleaner but might cause less performance but I don't think that's much. So your call.

timtofan:
> Approved. But just as an opinion, I think we might be able to call `listener.populateMetricsFoundStat` even when there is a loop which does something else as well. It's cleaner but might cause less performance but I don't think that's much. So your call.

yes I was thinking about it, guess it shouldn't really affect performance, lemme update

timtofan:
@emadolsky updated

emadolsky (approved):
Awesome! Thanks.

PR #522, opened 2023-02-07
Use intermediate chan to expedite gRPC render cache

emadolsky:
Before this commit, render responses were cached when the whole response was streamed to the client. This caused slowing requests waiting for that cache to be populated. Now, by using an intermediate channel when caching is enabled, we do not wait until the request is streamed.

### DIFF
--- a/carbonserver/render.go
+++ b/carbonserver/render.go
@@ -392,7 +392,7 @@ func getMetricGlobMapFromExpandedGlobs(expandedGlobs []globs) map[string]globs {
 	return metricGlobMap
 }
 
-func (listener *CarbonserverListener) prepareDataStream(ctx context.Context, format responseFormat, targets map[timeRange][]target, metricGlobMap map[string]globs, responseChan chan response) {
+func (listener *CarbonserverListener) prepareDataStream(ctx context.Context, format responseFormat, targets map[timeRange][]target, metricGlobMap map[string]globs, responseChan chan<- response) {
 	defer close(responseChan)
 	for tr, ts := range targets {
 		for _, metric := range ts {
@@ -622,7 +622,7 @@ func (listener *CarbonserverListener) fetchData(metric, pathExpression string, f
 	return multi, nil
 }
 
-func (listener *CarbonserverListener) streamMetrics(stream grpcv2.CarbonV2_RenderServer, responseChan chan response) (metricsFetched, valuesFetched, fetchSize int, err error) {
+func (listener *CarbonserverListener) streamMetrics(stream grpcv2.CarbonV2_RenderServer, responseChan <-chan response) (metricsFetched, valuesFetched, fetchSize int, err error) {
 	var metricAccessBatch []string
 	for renderResponse := range responseChan {
 		protoRes := renderResponse.proto2()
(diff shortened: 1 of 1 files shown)


carbonserver/render.go
```diff
@@ -744,33 +737,56 @@ func (listener *CarbonserverListener) Render(req *protov2.MultiFetchRequest, str
...
-			})
-		tle.CacheDuration = float64(time.Since(cacheT0)) / float64(time.Second)
-		if err == nil {
-			listener.prometheus.cacheRequest("query", fromCache)
-			if fromCache {
-				atomic.AddUint64(&listener.metrics.QueryCacheHit, 1)
-				cacheResponse := res.([]response)
+
+		item := listener.queryCache.getQueryItem(key, size, 60)
+		res, cacheOk := item.FetchOrLock()
```
timtofan: It wasn't clear for me what `cacheOk` means here, after reading FetchOrLock() code I think it means something like "gotValue". If so, maybe worth renaming the variable to make it more intuitive.

Or maybe even more intuitive is not to check bool but res == nil

carbonserver/render.go
```diff
@@ -744,33 +737,56 @@ func (listener *CarbonserverListener) Render(req *protov2.MultiFetchRequest, str
...
+			responseChan := make(chan response, 1000)
+			err = fetchMetricsFunc(responseChan)
+			if err != nil {
+				item.StoreAbort()
+				tle.CacheDuration = float64(time.Since(cacheT0)) / float64(time.Second)
+			} else {
 				go func() {
-					for _, r := range cacheResponse {
-						responseChan <- r
+					var fullResponse []response
```
timtofan: nit: maybe `responses` would be simpler, `full` phrasing could be a bit confusing, like if there is full then there must be partial which makes you wonder and check further, while `responses` is clear than it's just array of `response`s

`responses` is also good in that it suggests that's it's an array/slice

carbonserver/render.go
```diff
@@ -744,33 +737,56 @@ func (listener *CarbonserverListener) Render(req *protov2.MultiFetchRequest, str
...
+		case res != nil:
+			atomic.AddUint64(&listener.metrics.QueryCacheHit, 1)
+			go func() {
+				for _, r := range res.([]response) {
+					streamChan <- r
+				}
+				close(streamChan)
+			}()
+			fromCache = true
+			tle.CacheDuration = float64(time.Since(cacheT0)) / float64(time.Second)
```
timtofan: maybe this should be inside goroutine above? otherwise it can executed before goroutine finishes I think 🤔

carbonserver/render.go
```diff
@@ -711,9 +708,9 @@ func (listener *CarbonserverListener) Render(req *protov2.MultiFetchRequest, str
 	valuesFetched := 0
 	fetchSize := 0
 	// TODO: should chan buffer size be configurable?
-	responseChan := make(chan response, 1000)
+	streamChan := make(chan response, 10000)
```
timtofan: This variable is is used quite a lot and quite far from definition, so I think it's important for it to have intuitive name. streamChan naming doesn't give any intuition on what's inside, maybe something like `marshalledResponsesChan` would be more clear.

carbonserver/render.go
emadolsky: It's about whether the results were found in the cache or not.
So changed to `found`.

carbonserver/render.go
emadolsky: Changed to `responses`

carbonserver/render.go
emadolsky: Yeah makes sense. Changed.

carbonserver/render.go
emadolsky: Changed it to `responseChanToStream`. Because it is the channel which contains the responses which will be streamed.

emadolsky: BTW, They are not marshalled.

timtofan: ah indeed not marshalled, that was for http, cool thx

timtofan (approved):

PR #526, opened 2023-02-10
Limit streaming channel size dynamically for gRPC render

emadolsky:
if number of files in the expanded globs is smaller than that maximum, it makes sense to initiate chan with that size instead of maximum.

### DIFF
--- a/carbonserver/render.go
+++ b/carbonserver/render.go
@@ -388,6 +388,14 @@ func getMetricGlobMapFromExpandedGlobs(expandedGlobs []globs) map[string]globs {
 	return metricGlobMap
 }
 
+func getStreamingChannelSize(filesCount int) int {
+	channelSize := 100
+	if filesCount < channelSize {
+		channelSize = filesCount
+	}
+	return channelSize
+}
+
 func (listener *CarbonserverListener) prepareDataStream(ctx context.Context, format responseFormat, targets map[timeRange][]target, metricGlobMap map[string]globs, responseChan chan<- response) {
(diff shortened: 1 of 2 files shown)


carbon/app.go
```diff
@@ -510,6 +510,12 @@ func (app *App) Start(version string) (err error) {
 		carbonserver.SetNoServiceWhenIndexIsNotReady(conf.Carbonserver.NoServiceWhenIndexIsNotReady)
 		carbonserver.SetRenderTraceLoggingEnabled(conf.Carbonserver.RenderTraceLoggingEnabled)
 
+		if conf.Carbonserver.MaxStreamingChannelSize == nil || *conf.Carbonserver.MaxStreamingChannelSize < 0 {
+			carbonserver.SetMaxStreamingChannelSize(100)
```
azhiltsov: Default configs are in carbon/config.go
I think it make sense to move defaults there

emadolsky: We had this discussion and decided to remove the whole parameter.

grzkv (approved):

azhiltsov (approved):

carbonserver/render.go
```diff
@@ -388,6 +388,14 @@ func getMetricGlobMapFromExpandedGlobs(expandedGlobs []globs) map[string]globs {
 	return metricGlobMap
 }
 
+func getStreamingChannelSize(filesCount int) int {
+	channelSize := 100
```
timtofan, line 392: maybe `maxChannelSize`

PR #534, opened 2023-02-14
added metric for ooo lag for each datapoint

auguzun:
we receive OOO writes in graphite storage, we need to understand how far they are in the past in order to suggest possible solutions for OOO for cwhisper. In this PR I added prometheus histogram metric for OOO lag for each datapoint.
 Tested on prod box, I didn't notice any performance impact

### DIFF
--- a/persister/whisper.go
+++ b/persister/whisper.go
@@ -88,9 +88,9 @@ type Whisper struct {
 
 // NewWhisper create instance of Whisper
 type whisperPrometheus struct {
-	enabled              bool
-	outOfOrderWritesLags prometheus.Histogram
-	outOfOrderWritesLag  func(time.Duration)
+	enabled             bool
+	outOfOrderWriteLags prometheus.Histogram
+	outOfOrderWriteLag  func(time.Duration)
 }
 
 func NewWhisper(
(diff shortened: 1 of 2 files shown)


azhiltsov (approved):

persister/whisper.go
```diff
@@ -188,7 +212,16 @@ func fnv32(key string) uint32 {
 	}
 	return hash
 }
-
+func (p *Whisper) registerOutOfOrderLag(points []*whisper.TimeSeriesPoint) {
```
timtofan: maybe `registerOutOfOrderLags` since it saves multiple timings?

persister/whisper.go
```diff
@@ -3,6 +3,7 @@ package persister
 import (
 	"errors"
 	"fmt"
+	prom "github.com/prometheus/client_golang/prometheus"
```
timtofan: linter warning
> File is not `goimports`-ed (goimports)

timtofan: in general I think renaming imports is frowned upon since it adds extra complexity to remember not the package name but import name, maybe it's really needed in this case though, not sure

deniszh: I think it just want to separate import for named importing. But it's just easier to [include goimports to gofmt](https://pkg.go.dev/golang.org/x/tools/cmd/goimports).

persister/whisper.go
```diff
@@ -200,6 +234,7 @@ func (p *Whisper) updateMany(w *whisper.Whisper, path string, points []*whisper.
 	}()
 
 	// start = time.Now()
+	p.registerOutOfOrderLags(points)
```
deniszh: is it good idea to do that syncronously? maybe better put to goroutine?

deniszh (approved):

persister/whisper.go
auguzun: I did not notice any performance impact during testing, do you think we might have it in some cases?

persister/whisper.go
auguzun: thanks, removed renaming and fixed linter

persister/whisper.go
auguzun: fixed

persister/whisper.go
```diff
@@ -77,14 +79,20 @@ type Whisper struct {
...
 	// blockThrottleNs        uint64 // sum ns counter
 	// blockQueueGetNs        uint64 // sum ns counter
 	// blockAvoidConcurrentNs uint64 // sum ns counter
 	// blockUpdateManyNs      uint64 // sum ns counter
 }
 
 // NewWhisper create instance of Whisper
+type whisperPrometheus struct {
+	enabled              bool
+	outOfOrderWritesLags prometheus.Histogram
```
timtofan: I think double plurals are harder to read, `outOfOrderWriteLags` might be more digestable, lemme see if I can find an article on that

timtofan: https://english.stackexchange.com/a/258117

persister/whisper.go
```diff
@@ -188,7 +213,16 @@ func fnv32(key string) uint32 {
 	}
 	return hash
 }
-
+func (p *Whisper) registerOutOfOrderLags(points []*whisper.TimeSeriesPoint) {
```
timtofan: guess we could keep name similar to prom metric, ie `registerOutOfOrderWriteLags`

persister/whisper.go
auguzun: tested on the most loaded box, no issues there

persister/whisper.go
auguzun: thanks, fixed

persister/whisper.go
auguzun: thanks, fixed

PR #541, opened 2023-02-20
Upgrade golang.org/x/net

deniszh:
Because of dependabot https://github.com/go-graphite/go-carbon/security/dependabot/6


deniszh:
Very strange error. Well, maybe it's better to address that later. 🤷

deniszh:
Probably I did that completely wrong, we just need to wait until it would be patched in supported go versions.

PR #559, opened 2023-06-23
Massive go deps upgrade

deniszh:


deniszh:
Welp, that breaks everything beyond golang v1.21, probably not worth it :D

PR #612, opened 2024-08-06
Make throughput quota config per minute

emadolsky:
Previously, it was configured for a full usage frequency interval. This caused configuration change on throughput values when the frequency was changed. We prefer to be able to change the frequency without changing any other value and expect the same results.


deniszh (approved):

grlvrl (approved):
LGTM

PR #692, opened 2025-02-14
Docker: build mulitarchitecture images

RoEdAl:
Thanks to [docker/build-push-action@v6](http://github.com/docker/build-push-action) action now it's easy to build multiarchitecture Docker images. So, I decided to modify `docker-ghcrio.yml` workflow and build (publish) Docker images for **ARM64** (`linux/arm64`) and **ARM** (`linux/arm/v7`,`linux/arm/v6`) architectures. `Dockerfile` was also improved a bit.

Related PR: https://github.com/go-graphite/carbonapi/pull/858

### DIFF
--- a/.github/workflows/docker-ghcrio.yml
+++ b/.github/workflows/docker-ghcrio.yml
@@ -1,19 +1,27 @@
-name: Upload Docker images to ghcr.io
+name: Upload Docker images to GitHub Container Registry (ghcr.io)
 
 on:
   push:
-    branches: [ master ]
+    branches: [ master, main ]
     tags: [ 'v*' ]
   pull_request:
-    branches: [ master ]
+    branches: [ master, main ]
 
 jobs:
   docker:
-    name: Build image
+    name: Build images
     runs-on: ubuntu-latest
     steps:
       - name: Check out code
         uses: actions/checkout@v4
+      - name: Set up QEMU
+        if: github.event_name != 'pull_request'
+        uses: docker/setup-qemu-action@v3
+        with:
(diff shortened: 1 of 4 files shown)


deniszh:
Hi @RoEdAl 

Thanks, I like your idea, but:
1. tests are broken, could you please check?
2. image build run for 30m and not even finished, maybe it's stuck because of broken tests. But if not, maybe it's better to stick to amd64 and linux/arm64 only ?

RoEdAl:
1. `Dockerfile` was fixed. Please run tests again.
2. Building all images takes about 45m. This is normal because non-x86 architectures uses QEMU which is rather slow. Furthermore, tests are performed on every platform (architecture).

deniszh:
45 min is too long, I'm afraid. We need to do that somehow asynchronously, for not waiting 45 minutes to merge PR

RoEdAl:
Maybe `docker-ghcrio.yml` workflow triggered by PR should build only one `linux/amd64` Docker image?

deniszh:
Make sense. We can build everything only for releases.

deniszh:
PS: tests still not fixed :(
You can just test them using "make image"

RoEdAl:
The problem is that `make image` command works fine on my computer.
My next try is to execute `make clean` before `make` inside Docker container so please run tests again.

RoEdAl:
Updating workflow. Now only `linux/amd64` image is build when triggered by PR.

deniszh:
Looks good, merging!

PR #896, opened 2026-07-23
perf(tcp): parallelize plaintext receiver processing

deniszh:
## Summary

  - Add configurable TCP plaintext workers (workers, default 1) to parallelize processing from a single connection.
  - Batch buffered reads to reduce channel overhead.
  - Drain complete and queued batches during graceful shutdown to prevent metric loss.
  - Update configuration docs and add concurrency, shutdown, and benchmark coverage.

## Results

  Benchmark improved from ~4.3M to ~8.6M points/sec with four workers (~2×). TCP race tests and all receiver tests pass.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,5 @@ sha256sum
 *.deb
 /local/
 .omc
+# Added by code-review-graph
+.code-review-graph/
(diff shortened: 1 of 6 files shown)


deniszh:
@copilot Fix the code for all comments in [this review thread](https://github.com/go-graphite/go-carbon/pull/896#pullrequestreview-4765330259).

When a review comment includes a suggested change, apply the suggestion exactly.

Do not make changes beyond what is described in the linked review thread.

